PR #15, opened 2018-04-04
Allow 6 cars per train on the Wooden RC's 6-seater trains

Gymnasiast:
'The Beast' on King's Island has this configuration, for example.


duncanspumpkin:
Wont this sort of modification get wiped out whenever we rebuild the objects when there is an update to the object exporter.

duncanspumpkin:
Will that not be a bit of a pain when flags get named / additional information gets added to an object?

IntelOrca:
Search and replace, as soon as custom objects get made we will then have to support old names forever even if as aliases.

PR #17, opened 2018-04-20
update will complete later

Omaranwa:
this is so hard for completing, why you did this?


objects/rct1/wall/rct1.wfw0.json
```diff
@@ -28,6 +28,7 @@
             "ja-JP": "木製のフェンス",
             "pl-PL": "Drewniany płot",
             "ru-RU": "Забор"
```
AaronVanGeffen: Missing comma

AaronVanGeffen:
> this is so hard for completing, why you did this?

These translations really aren't supposed to be done through the GitHub web interface. At the moment, we don't have any specific tooling for it either, though. Given we can't render Arabic properly yet anyway, I suggest you hold off translating until we do.

janisozaur:
> hold off translating until we do.

Or work on providing such tools.

PR #26, opened 2018-07-22
es-ES: Removed a bad translation line.

frenchiveruti:
The original english text was still on the car name, so I removed it in order to leave only the es-ES String.


objects/rct2/ride/rct2.obs2.json
```diff
@@ -130,7 +130,7 @@
             "en-GB": "Double-deck Cabin",
             "fr-FR": "Tour d'observation 2-ponts",
             "de-DE": "Doppeldecker-Aussichtsturm",
-            "es-ES": "Torre de Observación DobleDouble-deck Observation Tower",
+            "es-ES": "Torre de Observación Doble",
```
Gymnasiast: This looks like a translation for "Double-deck observation tower", not "Double-deck observation **cabin**".

frenchiveruti:
Ah... You're right, I didn't even try, ok, please deny the PR and I will come with the correct name that should be:

Cabina de observación doble piso.

frenchiveruti:
@Gymnasiast Now it should be an actual translation.

PR #27, opened 2018-08-08
Tooling for dumping and loading translations into JSON objects

AaronVanGeffen:
Currently, adding new strings to JSON objects is quite a pain for translators. In this PR, I would like to propose tooling to make life a little easier for them.

#### language_dump.py

This script outputs all existing strings for one particular language, falling back to en-GB if there are no translations yet. Example output:

```
$ ./language_dump.py -d out.json -l ja-JP > /dev/null
$ head out.json 
{
    "rct2.roof12": {
        "name": "屋根"
    },
    "rct2.tl1": {
        "name": "木"
    },
    "rct2.tt1": {
        "name": "ローマ寺院"
    },
```

A dump of all current en-GB strings may be found here for reference: https://pastebin.com/jLYjPyfL

#### language_load.py

This script does the reverse: it takes a JSON dump file and writes the new strings to the appropriate JSON object files. If the proposed string is identical to en-GB, it is omitted.

Caveat: the object files currently have non-standard indentation with regards to arrays. I have extended Python's JSON encoder to compensate, only to find we are inconsistent about it. This requires some more work, either way.

### DIFF
--- a/language_dump.py
+++ b/language_dump.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+
+import argparse
+import glob
+import json
+# from pprint import pprint
+
+# Command line arguments.
+parser = argparse.ArgumentParser(description='Dump translations for OpenRCT2\'s JSON objects.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('-o', '--objects', default="objects", help='JSON objects directory')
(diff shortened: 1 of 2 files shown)


Gymnasiast:
I now see I never reviewed this PR! I will when I have the time and opportunity (not at home atm).

IntelOrca:
They aren't inconsistant, it does have rules based on number of items and which field. An encoder that can just edit the changed properties rather than the whole document would be easier though.

PR #28, opened 2018-08-16
Rename fries to chips for UK spelling

Broxzier:
Reference: http://grammarist.com/usage/chips-vs-fries/

There was some talk in the Roller Coaster & Friends Discord server about whether is should be "Chip Stall" or "Chips Stall" (it's called Chip Hut in RCT1). Soemone with an English degree said there's not a definitive answer to which one is more correct.

Does this need to bump the version somewhere, or is that only for releases?

### DIFF
--- a/objects/rct2/ride/rct2.chpsh2.json
+++ b/objects/rct2/ride/rct2.chpsh2.json
@@ -19,7 +19,8 @@
     "images": ["$RCT2:OBJDATA/CHPSH2.DAT[0..6]"],
     "strings": {
         "name": {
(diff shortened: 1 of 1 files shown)


IntelOrca:
The reference article doesn't mention UK distinction between fries and chips which is usually that fries or french fries are long and thin where as chips are thick. Hence why the game has two stalls, one that sells thin and the other that sells thick.

I think chips is correct as the stall always sells a bag of chips per sale not just a single one.

Gymnasiast:
Hold off tagging, there is more I want to do with regards to objects soon.

PR #38, opened 2018-12-26
it-IT new station translations

LucaRed:
New it-IT translations as per OpenRCT2/Localisation#1625

I added my new strings in the language code alphabetical order, however not every file is currently fully ordered. I think that the alphabetical order makes sense because:
1) it's cleaner
2) you don't have to edit two lines to add the comma whenever you add a new translation (except if your language is the last one in the alphabetical order).


Gymnasiast:
I think those are very good reasons. :+1:

PR #52, opened 2019-10-27
Make English vehicle names more consistent

### LINKED ISSUE #50: Rename some low quality vehicle names/descriptions
A couple of vehicle names/descriptions are of low quality or don't make sense. Most vehicles have “train” or “car” in their name, but some vehicle names and descriptions are weird or just outright wrong.

Here I present the problematic strings (in `en-GB`) and suggest changes. Please post what you think of my suggestions and if they are OK.

1:
```
Small flower shaped dinghies powered by a centrally-mounted motor controlled by the passengers
	→ Source: rct2.tt.flwrpowr.json
	→ I thought dinghies are boats and need water, but this ride doesn't look like it has any water.
	→ Suggestion: Small flower shaped dodgems powered by a centrally-mounted motor controlled by the passengers
```

2:
```
Roller coaster cars in the shape of lizard carnival floats
Roller coaster cars in the shape of butterfly carnival floats
Roller coaster cars in the shape of frog carnival floats
	→ Source: rct2.ww.crnv*.json
	→ These are cars for car rides, not for roller coasters!
	→ Suggestion: Cars in the shape of lizard/butterfly/frog carnival floats
```

3:
```
Carnival Float Ride - Butterfly Cars
Carnival Float Ride - Frog Cars
Carnival Float Ride - Lizard Cars
	→ Source: rct2.ww.crnv*.json
	→ These vehicle names are a mouthful, let's shorten them!
	→ Suggestion: Carnival Butterfly/Frog/Lizard Cars
```

4: These rollercoaster trains/cars should be renamed to become a proper vehicle name:
```
Bullet Coaster
	Source: rct2.ww.bullet.json
	Suggestion: Bullet Trains
Dragon Coaster
	Source: rct2.ww.dragon.json
	Suggestion: Dragon Trains
Limousine Rollercoaster
	Source: rct2.ww.cadillac.json
	Suggestion: Limousine Trains
Seals Coaster
	Source: rct2.ww.seals.json
	Suggestion: Seal Trains
```

5: Also consider renaming these:
```
Jaguar Ride
	Source: rct2.ww.jaguarrd.json
	Suggestion: Jaguar Cars
Sloth Ride
	Source: rct2.ww.sloth.json
	Suggestion: Sloth Trains
Chinese Dragonhead Ride
	Source: rct2.ww.sloth.json
	Suggestion: Chinese Dragonhead Dodgem Cars
Flower Power Ride
	Source: rct2.tt.flwrpowr.json
	Suggestion: Flower Power Dodgem Cars
```

6:
```
Raptor Racers Coaster
	Source: rct2.tt.raptorxx.json
	Suggestion: “Raptor Racers” or “Racing Raptors”
```

Wuzzy2:
Fixes #50.

Vehicles are renamed according to some rules:

* Vehicles must have a name that describes the vehicle itself rather than the ride/rollercoaster as a whole
* Exception: Dodgems, Flying Saucers and all their variants introduced in WW/TT. The term “Dodgems” instead of “Dodgem Cars” is used because all variants appear as unique ride in the ride list, using the vehicle name as the ride name
* Use plural form unless the ride can only have 1 vehicle
* Consistent capitalization
* Problems reported in #50 are fixed

Affects only `en-GB`.

### DIFF
--- a/objects/rct2tt/ride/rct2.tt.ganstrcr.json
+++ b/objects/rct2tt/ride/rct2.tt.ganstrcr.json
@@ -108,7 +108,7 @@
             "ru-RU": "Горки гангстеров"
         },
         "description": {
-            "en-GB": "1920s Themed Gangster cars are accelerated out of the station by linear induction motors to speed through twisting inversions",
+            "en-GB": "1920s-themed gangster cars are accelerated out of the station by linear induction motors to speed through twisting inversions",
             "fr-FR": "Train sur le thème des gangsters des années folles propulsé hors de la station par des moteurs à induction linéaire et fonçant dans des inversions renversantes",
             "de-DE": "Die Gangster-Fluchtwagen aus den Zwanzigern werden aus der Station heraus mit Linearmotoren beschleunigt und rasen durch verdrehte Loopings.",
             "es-ES": "Los coches, ambientados como de gángsteres de los años veinte tienen que ser acelerados por motores de inducción lineal, para coger velocidad para los giros en inversión.",
--- a/objects/rct2tt/ride/rct2.tt.policecr.json
+++ b/objects/rct2tt/ride/rct2.tt.policecr.json
@@ -117,7 +117,7 @@
             "ru-RU": "Полиция"
         },
         "description": {
-            "en-GB": "1960s Themed Police cars run on wooden tracks, turning around on special reversing sections",
+            "en-GB": "1960s-themed police cars run on wooden tracks, turning around on special reversing sections",
             "fr-FR": "Voitures de police des années 60 dévalant une voie en bois, tournant à des sections inverseuses spéciales",
             "de-DE": "Polizeiautos aus den 60ern fahren über eine Holzstrecke und drehen an speziellen Wendeabschnitten.",
             "es-ES": "Coches decorados como de la policía en los años sesenta que circulan por guías de madera, girando completamente en secciones de inversión especiales.",
--- a/objects/rct2ww/ride/rct2.ww.coffeecu.json
+++ b/objects/rct2ww/ride/rct2.ww.coffeecu.json
@@ -413,7 +413,7 @@
             "pt-BR": "Xícaras Giratórias"
         },
         "description": {
-            "en-GB": "Riders ride in pairs of seats rotating around the ends of three long rotating arms",
+            "en-GB": "Riders ride in pairs of themed seats rotating around a large animated coffee grinder",
             "fr-FR": "Les passagers sont assis par deux et tournent autour de l'extrémité de trois longs bras rotatifs",
             "de-DE": "Die Fahrgäste nehmen auf rotierenden Sitzpaaren Platz, die sich an den Enden dreier langer kreisender Arme befinden",
             "es-ES": "Los pasajeros montan en parejas de dos asientos que giran alrededor de los extremos de tres largos brazos rotatorios",
--- a/objects/rct2ww/ride/rct2.ww.italypor.json
+++ b/objects/rct2ww/ride/rct2.ww.italypor.json
@@ -413,7 +413,7 @@
             "pt-BR": "Polícia Italiana"
         },
         "description": {
-            "en-GB": "Riders ride in pairs of seats rotating around the ends of three long rotating arms",
+            "en-GB": "Riders ride in pairs in Italian police car replicas and rotate in circles",
             "fr-FR": "Les passagers sont assis par deux et tournent autour de l'extrémité de trois longs bras rotatifs",
             "de-DE": "Die Fahrgäste nehmen auf rotierenden Sitzpaaren Platz, die sich an den Enden dreier langer kreisender Arme befinden",
             "es-ES": "Los pasajeros viajan en parejas de asientos que rotan alrededor de los extremos de tres largos brazos giratorios",
--- a/objects/rct2/ride/rct2.arrx.json
+++ b/objects/rct2/ride/rct2.arrx.json
@@ -86,7 +86,7 @@
     "images": ["$RCT2:OBJDATA/ARRX.DAT[0..28226]"],
     "strings": {
         "name": {
-            "en-GB": "Multi-Dimension Coaster Train",
+            "en-GB": "Multi-Dimension Coaster Trains",
             "fr-FR": "Train multidimension",
             "de-DE": "Mehrdimensionaler-Achterbahnzug",
             "es-ES": "Montaña Rusa Multi-Dimensión",
--- a/objects/rct2/ride/rct2.intst.json
+++ b/objects/rct2/ride/rct2.intst.json
@@ -102,7 +102,7 @@
             "ru-RU": "Гига-Горки"
         },
         "description": {
-            "en-GB": "A giant steel roller coaster capable of smooth drops and hills of over 300ft",
+            "en-GB": "Roller coaster trains for the Giga Coaster, capable of smooth drops",
             "fr-FR": "Montagnes russes géantes proposant des chutes régulières depuis des collines dépassant les 100 mètres d'altitude",
             "de-DE": "Eine gigantische Stahlachterbahn mit gemäßigten Gefällen und Anstiegen von über 90 Meter",
             "es-ES": "Una montaña rusa gigante capaz de caídas suaves y subidas de más de 300 pies",
--- a/objects/rct2/ride/rct2.pmt1.json
+++ b/objects/rct2/ride/rct2.pmt1.json
@@ -83,7 +83,7 @@
     "images": ["$RCT2:OBJDATA/PMT1.DAT[0..1822]"],
     "strings": {
         "name": {
-            "en-GB": "Powered mine train",
+            "en-GB": "Powered Mine Trains",
             "fr-FR": "Mineur",
             "de-DE": "Minenbahn",
             "es-ES": "Trenes Mineros",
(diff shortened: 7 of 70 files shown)


objects/rct2/ride/rct2.bboat.json
```diff
@@ -65,7 +65,7 @@
             "ru-RU": "Лодочки"
         },
         "description": {
-            "en-GB": "Small circular dinghies powered by a centrally-mounted motor controlled by the passengers",
+            "en-GB": "Small circular dodgems powered by a centrally-mounted motor controlled by the passengers",
```
duncanspumpkin: This isn't correct though you would call them dinghies and not dodgems

objects/rct2tt/ride/rct2.tt.cyclopsx.json
```diff
@@ -39,7 +39,7 @@
     "images": ["$RCT2:OBJDATA/CYCLOPSX.DAT[0..66]"],
     "strings": {
         "name": {
-            "en-GB": "Cyclops Ride",
+            "en-GB": "Cyclops Dodgems",
```
Gymnasiast: They're not dodgems. They're flying saucers.

Wuzzy2:
Well, why are the flower power thingies called “dinghies” then? What do you mean by this word?

Wikipedia tells me a dinghy is a boat: https://en.wikipedia.org/wiki/Dinghy

Which is obviously not the case here. No water. So it doesn't make sense to me.

About cyclops: How do you know they are flying saucers? Sure, the base ride is *called* flying saucers, but that doesn't mean the cyclops eyes are floating, too. Graphically, they look like vehicles on wheels. Something which the description also suggests:

> Riders drive the Eye of a Giant Cyclops

You can't “drive” something that is hovering above the ground. It also sounds weird to say that these giant things actually hover above the ground like the flying saucers.

Or is “Cyclops Ride” actually the correct term? Do I have to read the word “Ride” as in “colloqial term for car” here? Or is “ride” referring to the whole building.

Sorry, not a native English speaker here. :(

Wuzzy2:
I have pushed an update now.

I have reverted the `bboat` change. Now I see the error here. I thought this was for the flower power ride, but it's actually for the bumper boats … I was completely confused.

I have changed “Cyclops Ride” to “Cyclops Cars”.

The reason why I find it odd to call the Cyclops things “flying saucer” is because these look very big, clunky and therefore heavy. I find it hard to believe they actually float above the ground like flying saucer. The name “Cyclops Flying Saucers” would be weird. Also, the current description of this vehicle doesn't say these vehicles are floating as well.

If you still object, I will revert. But I dislike “Cyclops Ride” as being too generic of a term and also not be in plural like most vehicle names are.

Wuzzy2:
Major update!

I have now added many changes to the vehicle descriptions as well, as many of these were descriptions for the ride rather than the vehicle.

Needs a fresh review (sorry).

objects/rct2tt/ride/rct2.tt.cyclopsx.json
```diff
@@ -39,7 +39,7 @@
     "images": ["$RCT2:OBJDATA/CYCLOPSX.DAT[0..66]"],
     "strings": {
         "name": {
-            "en-GB": "Cyclops Ride",
+            "en-GB": "Cyclops Cars",
```
Gymnasiast: These rides are listed as independent rides in the New Ride window, hence the `Cyclops Ride` name. Not sure if it should be changed.

Wuzzy2:
It seems the problem runs deeper.

The problem is that the same string is used for both vehicle and ride at the same time. Which is not the case for the other rides. This seems like a strange exception.

If we use “Cyclops Ride”, then the name is correct for the ride but odd for the vehicle.
If we use “Cyclops Cars”, then the name is correct (I hope) for the vehicle, but completely wrong for the ride. And it's not only the name. It's the description, too. The description has to match for the ride, not the vehicle. Which is very weird because most JSONs in the `ride` directory are for vehicles.

The real solution would be to use separate strings for both ride and vehicle. The vehicle could be read from JSON while the ride name is read from vanilla RCT2 (I hope this is possible).

I agree it's more convenient to have the different car types at least for dodgems and flying saucer-type listed as different rides. It's an exception to the rule atm, but in this case it makes sense. Fewer clicks. This can stay.

AAAAAAAaaaaaaaanyway …

I agree the name “Cyclops Cars” as of now doesn't make sense in the ride window, therefore, I have now reverted this to “Cyclops Ride” in the latest force-push (no other change). But I see this as a temporary solution and I will soon post a new issue on the main repo.

----

BTW, I really love that OpenRCT2 has lately completely thrown away the brain-dead “one vehicle = one ride” nonsense from vanilla RCT2 with stuff like “ostrich roller coaster” which is actually just a new vehicle. This was so annoying and it was definitely the right choice to completely nuke it. :)
But I see this change still needs some work; a lot of the vehicle descriptions are still wrong/outdated as they describing the ride rather the vehicle. Hence, this PR. :)

objects/rct2tt/ride/rct2.tt.ganstrcr.json
```diff
@@ -108,7 +108,7 @@
             "ru-RU": "Горки гангстеров"
         },
         "description": {
-            "en-GB": "1920s Themed Gangster cars are accelerated out of the station by linear induction motors to speed through twisting inversions",
+            "en-GB": "1920s themed gangster cars are accelerated out of the station by linear induction motors to speed through twisting inversions",
```
Gymnasiast:
```suggestion
            "en-GB": "1920s-themed gangster cars are accelerated out of the station by linear induction motors to speed through twisting inversions",
```

objects/rct2tt/ride/rct2.tt.policecr.json
```diff
@@ -117,7 +117,7 @@
             "ru-RU": "Полиция"
         },
         "description": {
-            "en-GB": "1960s Themed Police cars run on wooden tracks, turning around on special reversing sections",
+            "en-GB": "1960s themed police cars run on wooden tracks, turning around on special reversing sections",
```
Gymnasiast:
```suggestion
            "en-GB": "1960s-themed police cars run on wooden tracks, turning around on special reversing sections",
```

objects/rct2ww/ride/rct2.ww.italypor.json
```diff
@@ -413,7 +413,7 @@
             "pt-BR": "Polícia Italiana"
         },
         "description": {
-            "en-GB": "Riders ride in pairs of seats rotating around the ends of three long rotating arms",
+            "en-GB": "Riders ride in pairs in Italian police car replicas and rotate in circles",
```
Gymnasiast: Since this is the description that appears in the New Ride menu, I think it should mention the three long rotating arms.

objects/rct2ww/ride/rct2.ww.dragdodg.json
```diff
@@ -39,7 +39,7 @@
     "images": ["$RCT2:OBJDATA/DRAGDODG.DAT[0..66]"],
     "strings": {
         "name": {
-            "en-GB": "Chinese Dragonhead Ride",
+            "en-GB": "Chinese Dragonhead Dodgems",
```
Gymnasiast: They're not dodgems, but flying saucers.

Wuzzy2:
@Gymnasiast:

> They're not dodgems, but flying saucers.

But how do you know? Just because the ride class is “flying saucers” does not mean the vehicles are hovering. This particular vehicle type, just like the cyclops eye, looks quite bulky and heavy. Yeah, so my same rationale as before …

Interestingly, the JSON file is called `rct2.ww.dragdodg.json`. `dodg` sounds like “dodgems”.

Note that there's another vehicle/ride type, the “Skidoo Dodgems” (`rct2.ww.skidoo.json`) but is of class “flying saucers”. This seems to be the original vanilla RCT2 WW name, not altered. So not even RCT2 was so strict about this. Even in RCT2 being the ride type of “flying saucers” does not mean the vehicles can never be dodgem cars. Which makes sense, actually. “Flying saucers” as ride type is basically just dodgems but without roof.

> Since this is the description that appears in the New Ride menu, I think it should mention the three long rotating arms.

This doesn't make sense because this ride simply does not have any arms. At least I don't see any.

I fixed the 2 hypen things.

Wuzzy2:
Latest force-push changes:

* Name of Flower Power Ride restored
* Description of Flower Power Ride changed
* Name of Chinese Dragonhead Ride restored

These changes were too controversial apparently. *shrug*
I'm not 100% happy because the vehicle includes the word “ride” but this actually seems to be a different issue, see https://github.com/OpenRCT2/OpenRCT2/issues/10194

This PR has dragged on too long for now, and I don't wanna delay it any further.

Wuzzy2:
@Gymnasiast: Is anything else wrong here?

Gymnasiast:
Can you reply to the line notes themselves next time? Makes it easier to keep track of changes/discussion.

Wuzzy2:
Unfortunately, I am not able to reply to the line notes themselves. But I can write normal comments.

Gymnasiast:
You must be mistaken. Every regular user can reply to line notes. You can add replies in the input fields marked `Reply...`.

objects/rct2ww/ride/rct2.ww.coffeecu.json
```diff
@@ -413,7 +413,7 @@
             "pt-BR": "Xícaras Giratórias"
         },
         "description": {
-            "en-GB": "Riders ride in pairs of seats rotating around the ends of three long rotating arms",
+            "en-GB": "Riders ride in pairs of themed seats rotating around a large animated coffee mill",
```
Gymnasiast: From what I gather, coffee grinder (rather than coffee mill) appears to be most used translation of Kaffeemühle.

```suggestion
            "en-GB": "Riders ride in pairs of themed seats rotating around a large animated coffee grinder",
```

Wuzzy2:
@Gymnasiast: Done.

PR #60, opened 2020-03-18
Add images for RCT1 toilets

Gymnasiast:
This includes a thumbnail that I created myself, which matches the original one as closely as possible. I created it so it doesn't have the built-in mask that RCT1 thumbnails have.


IntelOrca:
0.png appears to be an information kiosk.

PR #61, opened 2020-03-19
Add images for most panda stuff

Gymnasiast:
This extracts the panda scenery images. The scenery group itself and the trains are absent, since I ran into recolouring issues.


AaronVanGeffen:
Interesting test case for custom objects to be based upon as well.

PR #64, opened 2020-03-29
Surfing Trains fixes

Fred-1044276:
I have made changes to the Surfing Trains:

- Increased maxCarsPerTrain from 7 to 9, for parity with the Twister Trains and Stand-up Twister Trains in RCT1 and OpenRCT2.
- Changed the English description to be more consistent with the Stand-up Twister Trains' new description.

This is part of my bigger project to fix the Wacky Worlds and Time Twister vehicles, either to better their descriptions (e.g. simplification, making the descriptions more specific), adding hasShelter flags, increasing the maxCarsPerTrain value to be more on par with these vehicles' base counterparts from RCT2 and RCT1, or a combination of the three.

### DIFF
--- a/objects/rct2ww/ride/rct2.ww.surfbrdc.json
+++ b/objects/rct2ww/ride/rct2.ww.surfbrdc.json
@@ -108,7 +108,7 @@
             "ru-RU": "Серфинг Коастер"
         },
         "description": {
-            "en-GB": "A surfboard-themed train with shoulder restraints, in which the riders stand up",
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Thanks! I like your project, the expansions are really a mess. We have done a good number of fixes already, but there is evidently a whole lot more.

For the next PR, could you try to make the commit title a bit more specific? I have now changed the title to `Change max vehicles on Surfing Trains; fix description` in order to make it clear what it does.

PR #76, opened 2020-04-08
Improve language dump and load python scripts

tupaschoal:
1. Specify encoding on language_dump.py I/O
2. Specify encoding on language_load.py I/O
3. Only allow languages listed in Localisation for language scripts
4. language_dump.py: Only dump on missing translations
5. language_load.py: Only print when a translation is updated

I think the commit titles are pretty self explanatory, but I'll explain the overall goal here.
- 1 and 2 fix problems I had while running the script.
- 3 makes it safer to run and paves the way for running it for all languages
- 4 and 5 make it more terminal friendly in case we're running for all languages.

This is a first building block for the project I called "Ease Objects Translation", which we discussed a week ago or so.

### DIFF
--- a/language_load.py
+++ b/language_load.py
@@ -5,12 +5,16 @@ import glob
 import json
 import re
 
+supported_languages = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES", "fi-FI",\
+                       "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL", "pl-PL", "pt-BR",\
+                       "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
+
 # Command line arguments.
 parser = argparse.ArgumentParser(description='Imports translations into OpenRCT2\'s JSON objects.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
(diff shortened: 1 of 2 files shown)


language_load.py
```diff
@@ -38,49 +44,56 @@ def iterencode(self, o, _one_shot=False):
...
 
         if not fallback_language in data['strings'][string_key]:
-            print("No en-GB reference for " + data['id'] + " string '" + string_key + "' in dump file -- probably shouldn't exist; skipping")
+            if verbose:
+                print("No en-GB reference for " + data['id'] + " string '" + string_key + "' in dump file -- probably shouldn't exist; skipping")
             continue
 
         if not language_to_import in data['strings'][string_key]:
             if strings_by_object[data['id']][string_key] == data['strings'][string_key][fallback_language]:
                 # TODO: Is this desirable behaviour?
```
tupaschoal, line 76: I don't think it is, because leaving them out will lead to them being marked as incomplete by the dump script.

AaronVanGeffen: Hmm, I think I added this to prevent importing untranslated strings, e.g. when a language pack is at a stage where 90% of strings are untranslated. I agree it will make it harder to do stats, though. Something for another PR, then.

tupaschoal: I think it is problematic because some things have the same translation for some languages and not for others and if we go towards having different files for each language on Localisation, people might want to have it "repeated" just to acknowledge that it was taken care of.

PR #82, opened 2020-04-08
Extract language_dump.py logic into functions

tupaschoal:
~*This is in draft while #76 is not merged, as it builds on top of it*~

This extracts `language_dump.py` logic into functions and also conforms to `pylint`

### DIFF
--- a/language_dump.py
+++ b/language_dump.py
@@ -1,37 +1,41 @@
 #!/usr/bin/env python3
+"""
+Dump translations for OpenRCT2\'s JSON objects given a language and a file
+"""
 
 import argparse
 import glob
 import json
 # from pprint import pprint
 
-supported_languages = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES", "fi-FI",\
-                       "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL", "pl-PL", "pt-BR",\
-                       "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
+SUPPORTED_LANGUAGES = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES",\
+                       "fi-FI", "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL",\
+                       "pl-PL", "pt-BR", "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
 
-# Command line arguments.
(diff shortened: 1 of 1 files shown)


language_dump.py
```diff
@@ -1,37 +1,41 @@
 #!/usr/bin/env python3
+"""
+Dump translations for OpenRCT2\'s JSON objects given a language and a file
```
AaronVanGeffen, line 3: The apostrophe doesn't need to be escaped here, right?
```suggestion
Dump translations for OpenRCT2's JSON objects given a language and a file
```

language_dump.py
```diff
@@ -40,16 +44,36 @@
...
                 if verbose:
                     print("Found existing translation for " + data['id'])
-                strings_by_object[data['id']][string_key] = data['strings'][string_key][language_to_extract]
+                strings_by_object[data['id']][string_key] = \
+                    data['strings'][string_key][language_to_extract]
             elif fallback_language in data['strings'][string_key]:
-                print("No existing translation for " + data['id'] + " yet, using English")
-                strings_by_object[data['id']][string_key] = data['strings'][string_key][fallback_language]
+                print("No existing translation for " + data['id'] + \
+                      " yet, using English")
```
AaronVanGeffen, line 51: This might be a good opportunity to use f-strings.

language_dump.py
```diff
@@ -40,16 +44,36 @@
             if language_to_extract in data['strings'][string_key]:
                 if verbose:
                     print("Found existing translation for " + data['id'])
-                strings_by_object[data['id']][string_key] = data['strings'][string_key][language_to_extract]
+                strings_by_object[data['id']][string_key] = \
+                    data['strings'][string_key][language_to_extract]
```
AaronVanGeffen, line 48: This feels a little awkward due to the wrapping. Perhaps save the rvalue first with a local (alias) assignment.

language_dump.py
```diff
@@ -40,16 +44,36 @@
...
+                    data['strings'][string_key][fallback_language]
             else:
                 if verbose:
-                    print("No existing translation for " + data['id'] + " yet, but no English string either -- skipping")
-                # pprint(data)
+                    print("No existing translation for " + data['id'] + \
+                          " yet, but no English string either -- skipping")
+
+def dump_translations():
+    """ Dump translations for OpenRCT2\'s JSON objects """
```
AaronVanGeffen, line 60:
```suggestion
    """ Dump translations for OpenRCT2's JSON objects """
```

language_dump.py
tupaschoal: I didn't know about that, will do!

tupaschoal:
I solved the issues and also removed some hardcoded `"English"` where it should actually be the fallback language

tupaschoal:
I did!

PR #83, opened 2020-04-09
Extract language_load.py logic into functions

tupaschoal:
~*This is in draft while #76 is not merged, as it builds on top of it*~

This extracts `language_load.py` logic into functions and also conforms to `pylint`

### DIFF
--- a/language_load.py
+++ b/language_load.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-Imports translations into OpenRCT2\'s JSON objects given a language and an input file
+Imports translations into OpenRCT2's JSON objects given a language and an input file
 """
 
 import argparse
@@ -125,10 +125,16 @@ def update_translation(target_lang, ref_lang, verbose, filename, strings_by_obje
         previously_translated = translation_existed(target_lang, data['strings'][string_key])
         translation_needs_update = False
         if previously_translated:
+            translation = data['strings'][string_key][target_lang]
+            target_string = strings_by_object[obj_id][string_key]
             translation_needs_update = translation_changed(verbose, obj_id, string_key,\
-                                          data['strings'][string_key][target_lang],\
-                                          strings_by_object[obj_id][string_key])
-        if not previously_translated or translation_needs_update:
+                                          translation, target_string)
+        else:
+            translation = data['strings'][string_key][ref_lang]
+            target_string = strings_by_object[obj_id][string_key]
+            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
+                                          translation, target_string)
+        if translation_needs_update:
             print("Updating " + obj_id + " string '" + string_key + "'")
             data['strings'][string_key][target_lang] = strings_by_object[obj_id][string_key]
             updated = True
--- a/language_load.py
+++ b/language_load.py
@@ -1,99 +1,152 @@
 #!/usr/bin/env python3
+"""
+Imports translations into OpenRCT2\'s JSON objects given a language and an input file
+"""
 
 import argparse
 import glob
 import json
 import re
 
-supported_languages = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES", "fi-FI",\
-                       "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL", "pl-PL", "pt-BR",\
-                       "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
+SUPPORTED_LANGUAGES = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES",\
+                       "fi-FI", "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL",\
+                       "pl-PL", "pt-BR", "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
 
-# Command line arguments.
-parser = argparse.ArgumentParser(description='Imports translations into OpenRCT2\'s JSON objects.', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-parser.add_argument('-o', '--objects', default="objects", help='JSON objects directory')
-parser.add_argument('-f', '--fallback', default="en-GB", help='Fallback language to check against', choices=supported_languages)
-parser.add_argument('-i', '--input', help='Translation dump file to import from', required=True)
-parser.add_argument('-l', '--language', help='Language that is being translated, e.g. ja-JP', required=True, choices=supported_languages)
-parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Maximize information printed on screen')
-args = parser.parse_args()
-
-language_to_import = args.language
-fallback_language = args.fallback
-verbose = args.verbose
-
-in_file = open(args.input, encoding="utf8")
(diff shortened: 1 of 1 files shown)


language_load.py
```diff
@@ -1,99 +1,152 @@
...
+        previously_translated = translation_existed(target_lang, data['strings'][string_key])
+        translation_needs_update = False
+        if previously_translated:
+            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
+                                          data['strings'][string_key][target_lang],\
+                                          strings_by_object[obj_id][string_key])
+        if not previously_translated or translation_needs_update:
+            print("Updating " + obj_id + " string '" + string_key + "'")
+            data['strings'][string_key][target_lang] = strings_by_object[obj_id][string_key]
             updated = True
```
tupaschoal, line 134: I've changed the behavior here, removing that `TODO` and merging the two branches, I'll paste them here for refrence:
```python
        if not language_to_import in data['strings'][string_key]:
            if strings_by_object[data['id']][string_key] == data['strings'][string_key][fallback_language]:
                # TODO: Is this desirable behaviour?
                if verbose:
                    print("Translation for " + data['id'] + " string '" + string_key + "' is identical to en-GB -- skipping")
                continue

            if verbose:
                print("Adding " + data['id'] + " string '" + string_key + "'")
            data['strings'][string_key][language_to_import] = strings_by_object[data['id']][string_key]
            updated = True
        else:
            if strings_by_object[data['id']][string_key] == data['strings'][string_key][language_to_import]:
                if verbose:
                    print("Translation for " + data['id'] + " string '" + string_key + "' has not changed -- skipping")
                continue

            print("Updating " + data['id'] + " string '" + string_key + "'")
            data['strings'][string_key][language_to_import] = strings_by_object[data['id']][string_key]
```

language_load.py
```diff
@@ -1,99 +1,152 @@
...
-                continue
-
-            print("Updating " + data['id'] + " string '" + string_key + "'")
-            data['strings'][string_key][language_to_import] = strings_by_object[data['id']][string_key]
+        previously_translated = translation_existed(target_lang, data['strings'][string_key])
+        translation_needs_update = False
+        if previously_translated:
+            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
+                                          data['strings'][string_key][target_lang],\
+                                          strings_by_object[obj_id][string_key])
```
AaronVanGeffen, line 130: I appreciate the use of booleans and functions to improve the readability, however I think this could be improved. I'd prefer to not break expressions into several lines if we can avoid it. Perhaps something like:
```suggestion
            translation = data['strings'][string_key][target_lang]
            target_string = strings_by_object[obj_id][string_key]
            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
                                          translation, target_string)
```

language_load.py
AaronVanGeffen: This means we'll have no mechanism avoiding the importing of untranslated (English) strings, correct? I'm a bit torn on this; it seems a bit wasteful. Moreover, if the English string changes in the mean time, the 'translation' won't be automatically updated, right?

language_load.py
```diff
@@ -1,99 +1,152 @@
 #!/usr/bin/env python3
+"""
+Imports translations into OpenRCT2\'s JSON objects given a language and an input file
```
AaronVanGeffen, line 3: The apostrophe doesn't need to be escaped in a Python docstring, right?
```suggestion
Imports translations into OpenRCT2's JSON objects given a language and an input file
```

language_load.py
tupaschoal: I see your point, my concern is that it might be annoying for some translators to have the script reporting the translation as incomplete because it was not incorporated due to the fact that it was equal to English (assuming we have everything automatic).

At the same time, if we do the approach I coded, it means that the translations will automatically gain a field duplicating the english text, with no intention.

What if we add something like `"NEEDS_TRANSLATION:"` on the dumping of the strings that fell back to the original language. Then when reading, we can check if it was removed, if it was, it means that the translator acknowledged that this should be kept as English.

tupaschoal: (By the way, I reverted to the old behavior for now, so we can implement the new one according to our discussion)

language_load.py
```diff
@@ -125,10 +125,16 @@ def update_translation(target_lang, ref_lang, verbose, filename, strings_by_obje
...
-                                          data['strings'][string_key][target_lang],\
-                                          strings_by_object[obj_id][string_key])
-        if not previously_translated or translation_needs_update:
+                                          translation, target_string)
+        else:
+            translation = data['strings'][string_key][ref_lang]
+            target_string = strings_by_object[obj_id][string_key]
+            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
+                                          translation, target_string)
+        if translation_needs_update:
```
AaronVanGeffen: For what it's worth, I don't think this fully reverts to the old behaviour? Both branches (if/else) are identical, and the `not previously_translated ` is missing here.

language_load.py
AaronVanGeffen: Per our earlier discussions, I figured it might be something like that, yes. I'm not sure which is the better alternative, to be honest.

Perhaps the script can _leave_ 'English' strings if present, but not _add_ them if somehow identical? That way, we have an override, while the script won't unnecessarily add untranslated strings.

language_load.py
tupaschoal: Doesn't it? Do note that the `translation` object changes from `target_lang` to `ref_lang` (that is the only difference.

So the update is done if `translation_existed && translation_changed` (old translation vs new translation) or `!translation_existed && translation_changed` (ref translation vs new translation)

```python
        previously_translated = translation_existed(target_lang, data['strings'][string_key])
        translation_needs_update = False
        if previously_translated:
            translation = data['strings'][string_key][target_lang]
            target_string = strings_by_object[obj_id][string_key]
            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
                                          translation, target_string)
        else:
            translation = data['strings'][string_key][ref_lang]
            target_string = strings_by_object[obj_id][string_key]
            translation_needs_update = translation_changed(verbose, obj_id, string_key,\
                                          translation, target_string)
        if translation_needs_update:
            print("Updating " + obj_id + " string '" + string_key + "'")
            data['strings'][string_key][target_lang] = strings_by_object[obj_id][string_key]
            updated = True
```

language_load.py
tupaschoal: > Perhaps the script can leave 'English' strings if present, but not add them if somehow identical? 

Not sure I understood this difference. You mean it won't rewrite English strings that have already been pushed and are equal to the reference, but it won't add any that remains equal?

AaronVanGeffen: Yes, I think we mean the same thing.  To illustrate, let's say en-GB and nl-NL both use the string "Hamburger" for an object. Dumping nl-NL and reloading it would still leave this string intact. 

Now, let's suppose this string is missing in de-DE. Dumping de-DE would see the reference string from en-GB inserted in its place. However, what I'm proposing is that reloading the new de-DE file would then _not_ create a "Hamburger" string in de-DE.

language_load.py
AaronVanGeffen: Ah, you're right. In that case, I'd really prefer this difference to be more explicit, by only branching between either assignment to `translation`. The rest can indeed be shared.

tupaschoal: Perfect, I've simplified it a great deal in the last commit.

language_load.py
tupaschoal: I agree it shouldn't and the current diff doesn't create this new string, so I'm marking this as solved. I'll take this discussion to an issue so we can decide there later and if more people want to chime in :)

PR #84, opened 2020-04-10
Report language translation completeness on language_dump.py

tupaschoal:
~*This is in draft while #77 is not merged, as it builds on top of it*~

This reports the translation completeness when extracting a language with `language_dump.py`:
It's in the form of:
```pt-BR: 99.96% completeness```

### DIFF
--- a/language_dump.py
+++ b/language_dump.py
@@ -30,12 +30,14 @@ def extract_translations(language_to_extract, fallback_language, verbose, filena
                          strings_by_object):
     """ Read JSON file and extracts translations for a given language """
 
+    reference_str_count = 0
+    translated_str_count = 0
     with open(filename, encoding="utf8") as file:
         data = json.load(file)
 
         if not 'strings' in data:
             print(f"No strings in {data['id']}, skipping")
-            return
+            return (translated_str_count, reference_str_count)
 
         for string_key in data['strings']:
             if not data['id'] in strings_by_object:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: language_dump.py [hunk 1]
```diff
@@ -63,10 +69,18 @@ def dump_translations():
...
-        extract_translations(language_to_extract, fallback_language, verbose,\
-                             filename, strings_by_object)
+        obj_translations, ref_obj_translation =\
+            extract_translations(language_to_extract, fallback_language, verbose,\
+                                 filename, strings_by_object)
+        reference_str_count += ref_obj_translation
+        translated_str_count += obj_translations
+
+    print('{}: {}% completeness'.format(language_to_extract,
+                                        round(100*translated_str_count/reference_str_count, 2)))
```
AaronVanGeffen, line 83: The formatting here is a little awkward. My suggestion would be to combine local assignments with f-strings again.

tupaschoal: Right, I forgot to update when rebasing, done now.

language_dump.py
```diff
@@ -63,10 +69,18 @@ def dump_translations():
...
     for filename in glob.iglob(args.objects + '/**/*.json', recursive=True):
-        extract_translations(language_to_extract, fallback_language, verbose,\
-                             filename, strings_by_object)
+        obj_translations, ref_obj_translation =\
+            extract_translations(language_to_extract, fallback_language, verbose,\
+                                 filename, strings_by_object)
+        reference_str_count += ref_obj_translation
+        translated_str_count += obj_translations
+
+    translation_progress = round(100*translated_str_count/reference_str_count, 2)
```
AaronVanGeffen, line 82:
```suggestion
    translation_progress = round(100 * translated_str_count / reference_str_count, 2)
```

Are you using pylint? I'd imagine this kind of thing would be covered by it.

tupaschoal: I'm it actually doesn't suggest this.

PR #87, opened 2020-04-12
Allow language_dump.py on all languages at once

tupaschoal:
~*This is in draft while #86 is not merged, as it builds on top of it. I'll rebase when it is merged*~

This allows the script to dump all languages at once, including their progress.
image


tupaschoal:
I did, it's working great :D

PR #94, opened 2020-04-12
Clean up non-wall references to 'shōji'

AaronVanGeffen:
The shōji term only refers to the walls, not the roofs or fences. I therefore propose we fall back to this less-specific, but more accurate, terminology.

### DIFF
--- a/objects/rct2ww/scenery_wall/rct2.ww.wshogi15.json
+++ b/objects/rct2ww/scenery_wall/rct2.ww.wshogi15.json
@@ -14,7 +14,7 @@
     "strings": {
         "name": {
             "en-GB": "Shōji Wall",
-            "fr-FR": "Cloison shogi",
+            "fr-FR": "Cloison shōji",
             "de-DE": "Shōji-Wand",
             "es-ES": "Pared Shōji",
             "it-IT": "Muro di Shōji",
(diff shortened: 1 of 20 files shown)


objects/rct2ww/scenery_wall/rct2.ww.wshogi15.json
```diff
@@ -13,15 +13,15 @@
     "images": ["$RCT2:OBJDATA/WSHOGI15.DAT[0..5]"],
     "strings": {
         "name": {
-            "en-GB": "Shōji Wall",
-            "fr-FR": "Cloison shogi",
-            "de-DE": "Shōji-Wand",
+            "en-GB": "Japanese Fence",
+            "fr-FR": "Clôture japonais",
+            "de-DE": "Japanische Hecke",
```
Gymnasiast, line 18: That doesn't sound quite right to me. Shouldn't it be "Zaun"?

@Wuzzy2

AaronVanGeffen: If that’s the case, there are a few more instances of Hecke that ought to be cleaned up. (Which should be much easier, now, with the scripts…)

Wuzzy2:
“fence” = “Zaun” in German.

“Japanese Fence” = “Japanischer Zaun”.

I have checked all instances of “Hecke” in German, but they are already correct. “Hecke” is German and means “hedge”.

If you don't speak German well, you shouldn't touch the German translation IMO.

AaronVanGeffen:
Noted. I’ll amend the changes and leave it to you in the future.

Wuzzy2:
Yep, looks good to me.

PR #104, opened 2020-05-23
Add images for the Panda trains and Panda theming

Gymnasiast:


objects/official/ride/rct2.dlc.zpanda/object.json
```diff
@@ -0,0 +1,876 @@
...
+        "",
+        "",
+        "",
+        "",
+        "",
+        "",
+        "",
+        "",
+        "",
+        ""
```
tupaschoal, line 830: What do all of these mean?

Gymnasiast: Empty images that need to be there because the game expects them.

AaronVanGeffen: This seems very hacky, though. Would implementing an empty fallback on the cpp side be an option?

Gymnasiast: As far as I know, there is a fallback, but it results in error messages like "image table longer/shorter than expected".

Gymnasiast: OK, I have done a test. They are used for the invisible cars at the begin and end. If you remove them, you will get strange artefacts, especially when setting the colour for every car.

AaronVanGeffen: Thanks for checking and explaining.

It's a shame JSON doesn't support comments (at least not in the  standard), because this seems like a good occasion for some. Perhaps you could add a brief explanation to the commit message at least.

Gymnasiast: > It's a shame JSON doesn't support comments (at least not in the standard), because this seems like a good occasion for some.

I fully agree - especially for images. I ran into the same problem when porting RCT1 vehicles over - it would have been so much easier if I could just group them with a comment.

Issue #93, opened 2020-04-12
Improve language_load.py JSON encoder


tupaschoal:
When loading back the translations with `language_load.py` it will often change the formatting of the JSON, as well as the newly translated strings. I've pasted the outputs below, using different encoders:

[Original JSON file](https://pastebin.com/7pvNRDL1)
[Updated JSON file with default JSON encoder](https://pastebin.com/5z4DXDNW)
[Updated ´JSON file with less verbose encoder](https://pastebin.com/WMcdszDF)

My idea is to either stop using encoders or investigate and improve the "less verbose encoder" to make sure we don't change the formatting too much or at all.

The "less verbose version" is what language_load.py currently uses. It does a good job on the strings, but makes it too verbose for other things. We either have to fix/modify it, or make sure we use one that is better suited.

tupaschoal:
Closing as this was made obsolete by #119

PR #119, opened 2020-07-10
Object dump import patch cleaning tool

AaronVanGeffen:
The object string dumper and loader scripts introduced in #27 have greatly simplified the workflow for object translations. However, the barrier of entry for use by most translators is still too big. Recent work by @tupaschoal has been working towards making an automated workflow, bridging the Objects and Localisation repositories. However, the holdup has been that Python's method of serialising JSON differs in output from what is already present (#93). Rather than writing our own encoder, I propose the following solution. 

I have written a simple Python class, PatchCleaner, that uses the `unidiff` package to disregard changes that arise in patches due to the formatting changes mentioned. In short, this means it only picks up the changes we need. I have added a heuristic to pick up on cases where a trailing comma was added to a line with an irrelevant language. This needs some field testing, but in my brief testing it has worked well.

The workflow is still suboptimal -- and the script doesn't take language and patch file as its arguments yet:
```bash
./language_dump.py -l nl-NL -d nl-NL_dump.json
$EDITOR nl-NL_dump.json
./language_load.py -l nl-NL -i nl-NL_dump.json
git diff > patch.diff
git stash
./language_clean_patch.py | git apply    # no arguments yet, sorry
git commit -m 'Updated object strings for nl-NL'
```

Again, this removes any need for cherry-picking the diff manually. This cherry-picking gets particularly cumbersome for rides and large scenery objects. Indeed, with this script, this workflow can now be completely automated!

As an aside: personally, I like to order the dump files before working on translations. We can avoid complicating `language_dump.py` by using `jq` to this end:
```
jq 'to_entries | sort_by(.value."reference-name") | from_entries' \
    < nl-NL_dump.json \
    > nl-NL_dump.json
```

### DIFF
--- a/language_clean_patch.py
+++ b/language_clean_patch.py
@@ -3,6 +3,29 @@
 from unidiff import PatchSet
 from unidiff.constants import LINE_TYPE_EMPTY
 
+import argparse
+import os
+
+SUPPORTED_LANGUAGES = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES",\
+                       "fi-FI", "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL",\
+                       "pl-PL", "pt-BR", "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
+
+def dir_path(string):
+    """ Checks for a valid dir_path """
+    if os.path.isdir(string):
+        return string
+    raise NotADirectoryError(string)
+
+def get_arg_parser():
+    """ Command line arguments """
+    parser = argparse.ArgumentParser(description=\
+                                     'Cleans up a patch file to apply to the objects repository.',
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-p', '--patch', required=True, help='Path to the patch file to clean up.')
+    parser.add_argument('-l', '--language', required=True, choices=SUPPORTED_LANGUAGES,\
+                                help='Language that is being translated, e.g. ja-JP')
+    return parser
+
 class PatchCleaner:
     """
     Cleans a given unified diff such that only lines matching the language parameter are included,
@@ -99,5 +122,8 @@ class PatchCleaner:
 
 
 if __name__ == "__main__":
-    patch = PatchCleaner("patch.diff", "nl-NL")
+    parser = get_arg_parser()
+    args = parser.parse_args()
+
+    patch = PatchCleaner(args.patch, args.language)
     print(patch)
--- a/language_clean_patch.py
+++ b/language_clean_patch.py
@@ -54,7 +54,6 @@ class PatchCleaner:
     def clean_hunk(self, hunk):
         """ Cleans one particular hunk in the patch set, removing irrelevant lines. """
 
-        prev_line = None
         for k, line in enumerate(hunk):
             # Is this line modifying anything?
             if line.is_added or line.is_removed:
@@ -65,15 +64,7 @@ class PatchCleaner:
 
                 # Is this line a removal followed by an addition of the same language,
                 # in turn followed by the target language?
-                if line.is_removed and hunk[k + 1].is_added and \
-                    self.language in hunk[k + 2].value:
-                    # TODO: Check languages match -- good enough a heuristic for now.
-                    continue
-
-                # Indeed, is the previous line a removal and the next line an addition
-                # of the target language?
-                elif hunk[k - 1].is_removed and line.is_added and \
-                    self.language in hunk[k + 1].value:
+                if self.is_accommodating_change(hunk, k):
                     # TODO: Check languages match -- good enough a heuristic for now.
                     continue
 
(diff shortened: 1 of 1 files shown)


language_clean_patch.py
```diff
@@ -0,0 +1,89 @@
...
+        """ Cleans one particular hunk in the patch set, removing irrelevant lines. """
+
+        prev_line = None
+        for k, line in enumerate(hunk):
+            # Is this line modifying anything?
+            if line.is_added or line.is_removed:
+                # We're definitely keeping lines that have something to do
+                # with the target language.
+                if self.language in line.value:
+                    continue
```
tupaschoal, line 64: Doesn't the dump script dumps missing strings on the target language as equal to the reference language?

This will add them back (but I'm not sure how to circumvent this for now)

language_clean_patch.py
```diff
@@ -0,0 +1,89 @@
...
+                elif hunk[k - 1].is_removed and line.is_added and \
+                    self.language in hunk[k + 1].value:
+                    # TODO: Check languages match -- good enough a heuristic for now.
+                    continue
+
+                # Otherwise, exclude this change from the patch.
+                else:
+                    line.line_type = LINE_TYPE_EMPTY
+
+            prev_line = k
```
tupaschoal, line 84: You don't really use `prev_line`

language_clean_patch.py
```diff
@@ -0,0 +1,89 @@
...
+        for k, line in enumerate(hunk):
+            # Is this line modifying anything?
+            if line.is_added or line.is_removed:
+                # We're definitely keeping lines that have something to do
+                # with the target language.
+                if self.language in line.value:
+                    continue
+
+                # Is this line a removal followed by an addition of the same language,
+                # in turn followed by the target language?
```
tupaschoal, line 67: Just to confirm, is this is checking for:
```diff
-pt-BR: 'something'
+pt-BR: 'something',
+nl-NL:'translatedsometing'
```

Maybe we should add a comment of what the scenario is like, exactly

language_clean_patch.py
```diff
@@ -0,0 +1,89 @@
...
+                if line.is_removed and hunk[k + 1].is_added and \
+                    self.language in hunk[k + 2].value:
+                    # TODO: Check languages match -- good enough a heuristic for now.
+                    continue
+
+                # Indeed, is the previous line a removal and the next line an addition
+                # of the target language?
```
tupaschoal, line 74: This is the same case as above, but we're pointing to the middle line, instead of the first one, right?

language_clean_patch.py
```diff
@@ -0,0 +1,89 @@
...
+                elif hunk[k - 1].is_removed and line.is_added and \
+                    self.language in hunk[k + 1].value:
+                    # TODO: Check languages match -- good enough a heuristic for now.
+                    continue
+
+                # Otherwise, exclude this change from the patch.
+                else:
+                    line.line_type = LINE_TYPE_EMPTY
```
tupaschoal, line 82: Is this what is read for the `.added` and `.removed`?

language_clean_patch.py
AaronVanGeffen: Interesting thought, but due to how the load script works, they wouldn't show up as an addition in the diff right? In case the load script decides to remove a line, this won't change the line status from removal to addition.

language_clean_patch.py
AaronVanGeffen: You're right! I ended up going for `hunk[k + 1]` and `hunk[k + 2]` instead. I'll remove the two references to `prev_line`.

language_clean_patch.py
AaronVanGeffen: That's correct. Perhaps a concrete example in the comments would be good, analogous to the one you just posted?

language_clean_patch.py
AaronVanGeffen: Correct. I thought about merging the two cases, but I believe it's clearer this way. What do you think?

language_clean_patch.py
AaronVanGeffen: Yes, indeed:
https://github.com/matiasb/python-unidiff/blob/286826e54f17dfa9c3d4873be575bcd83674007b/unidiff/patch.py#L98-L104

language_clean_patch.py
tupaschoal: Ok, if you say so. Just good to keep in mind in case it happens

language_clean_patch.py
tupaschoal: I guess it doesn't hurt to have the example

language_clean_patch.py
tupaschoal: I think I would make a method that does both and call it once, something like:
```python
if adding_new_item_to_list(line, hunk, self.language):
    continue
```

language_clean_patch.py
tupaschoal: Thanks for the confirmation :)

language_clean_patch.py
```diff
@@ -0,0 +1,129 @@
...
+
+from unidiff import PatchSet
+from unidiff.constants import LINE_TYPE_EMPTY
+
+import argparse
+import os
+
+SUPPORTED_LANGUAGES = ["ar-EG", "ca-ES", "cs-CZ", "da-DK", "de-DE", "en-GB", "en-US", "es-ES",\
+                       "fi-FI", "fr-FR", "hu-HU", "it-IT", "ja-JP", "ko-KR", "nb-NO", "nl-NL",\
+                       "pl-PL", "pt-BR", "ru-RU", "sv-SE", "tr-TR", "zh-CN", "zh-TW"]
```
tupaschoal, line 11: In the future we should probably have all scripts source the languages from a single place

AaronVanGeffen: Yes, definitely. I was also thinking we should probably move the language*.py tools into a Python package of some sort. Another item for the to do list…

PR #124, opened 2020-08-28
Add music objects for CUSTOM1 and CUSTOM2.WAV

IntelOrca:
Can be removed when new save format is released.


Gymnasiast:
> Can be removed when new save format is released.

What will we do if the user attempt to load a save that uses them?

IntelOrca:
I don't know, we can think about that around the time.

PR #125, opened 2020-09-25
Add shelter to Wacky Worlds Submarines

Fred-1044276:
Although the regular Submarine Ride vehicles were given the hasShelter property in the OpenRCT2 0.2.3 release, the Hippo Submarines and Killer Whale Submarines vehicles from Wacky Worlds were not given said properties. This PR fixes that by adding hasShelter properties to both the Hippo and Killer Whale vehicles for the Submarine Ride. It also adds a minor fix to the Killer Whale Submarines' English capacity description by replacing "car" with "submarine" for consistency with other Submarine Ride vehicles.

### DIFF
--- a/objects/rct2ww/ride/rct2.ww.hipporid.json
+++ b/objects/rct2ww/ride/rct2.ww.hipporid.json
@@ -11,6 +11,7 @@
             "gentle",
             "water"
         ],
+        "hasShelter": true,
         "noCollisionCrashes": true,
         "carColours": [
             [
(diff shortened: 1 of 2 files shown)


Gymnasiast (changes requested):
I have asked you several times not to mix unrelated properties. This PR should have two commits: one that adds the shelter property for both objects, and one that fixes the description.

Previously, I have fixed your PRs. I am not going to do that this time. This is at least the third time this has happened, so it looks like the message is not getting through.

AaronVanGeffen:
Thanks for the PR. I appreciate that you've made a smaller PR this time — and I think these are good changes!

I know it seems trivial for a PR this size. For various reasons, we really want to keep changes to logic separate from changes to strings. Could you amend this PR to remove the string change, and open another one for it?

AaronVanGeffen (approved):

PR #126, opened 2020-09-25
Add Asian Gong to Asia Themeing scenery group

Fred-1044276:
The Asian Gong small object (ORIEGONG) would always show up in the Miscellaneous Objects group, even if the Asia Themeing scenery group was available with all of its listed objects available. This is an issue that originated in RCT2 vanilla and has persisted in RCT Classic and OpenRCT2 until now. This PR fixes the issue by adding the Asian Gong object to the Asia Themeing scenery group. Note that any scenarios using the Asia Themeing that don't already have the Asian Gong object available in the scenario (e.g. Asia - Maharaja Palace) will still need to have said object re-added manually to the scenario with the Object selection feature during gameplay.

### DIFF
--- a/objects/rct2ww/scenery_group/rct2.ww.scgasia.json
+++ b/objects/rct2ww/scenery_group/rct2.ww.scgasia.json
@@ -39,6 +39,7 @@
             "GOODSAM",
             "GWOCTUR1",
             "GWOCTUR2",
+            "ORIEGONG",
             "TALLLAN1",
             "TALLLAN2",
             "TERRARMY",


Gymnasiast:
The gong is part of the Oriental theming that is part of vanilla. That theme contains several other objects that will end up in the Misc tab if selected without the Oriental theming present. What makes the gong special enough in this regard to warrant including it in the WW Asia theming?

AaronVanGeffen:
@Gymnasiast I'm not too versed in the Wacky Worlds expansions, but I'd say file the others into their relevant categories as well.

Gymnasiast:
Basically:
- Vanilla has Oriental theming.
- Wacky World has Asia theming. Obviously, there is some overlap.
- Some WW Asian scenarios use the going without including the Oriental theming tab, causing the gong to appear on the Misc tab.

AaronVanGeffen:
Hmm. Ideally, I would like the miscellaneous tab to not show up in vanilla scenarios. I realise the TT and WW scenarios are a bit of a mixed bag on this front, but if we can fix this with relative ease, I think we should. Perhaps by loading the scenery group object, even the scenario doesn't explicitly ask for it. That kind of change is beyond the scope of this PR, and indeed this repo, though.

Any objection to merging this as-is?

Gymnasiast:
> Any objection to merging this as-is?

Not convinced this object should be treated differently from all the other Oriental objects.

Gymnasiast:
Closing since OP has neglected to answer for a month.

Fred-1044276:
Do you really have to keep being a dumb asshole and close all my PRs over a tiny mistake? Shows what kind of a person you really are: unhelpful and bossy. This project would be better off without you and your authoritarian demeanors!

duncanspumpkin:
There is absolutely no reason to start name calling. @Gymnasiast was wanting further discussion as to why you thought the change should happen. You did not provide further discussion and the pr was then closed. When things come to style and layout there is always going to be dispute that is a fact of life. You shouldn't take this personally and neither should @Gymnasiast. Remember everyone here is a volunteer from yourself and up to Ted. Would you want to work on something if all you got was abuse, of course not. So do not start handing out abuse. You have made changes in the past that have been agreed with and merged and we do appreciate the work you put into this. So please keep things friendly and try to understand things from someone else's perspective.

Gymnasiast:
Unbelievable. I have put in a ton of work to fix/amend/split up your PRs in the past to make sure they can be merged and your work wasn't lost. I also cut you a lot of slack despite ignoring several requests not to bundle unrelated stuff in one PR. This is also the _first_ time a PR of you was closed without merging, no matter your instance I close "all your PRs". Not sure how that makes me an asshole.

PR #135, opened 2020-12-22
raptor train

spacek531:
Making the pull request now. Having some problems getting OpenRCT2 to recognize the object


spacek531:
I created a junction from `openrct2/objects` to the folder and OpenRCT2 does not load the object. Using `OpenRCT2.com` does not provide any console logs regarding the object. Not sure what's going on.

objects/official/ride/official.rmct2/object.json
```diff
@@ -0,0 +1,3470 @@
+{
+    "id": "official.rmct2",
+    "authors": ["spacek531"],
+    "version": "1.0",
+    "originalId": "00000000|RMCT2   |00003000",
+    "objectType": "ride",
+    "properties": {
+        "type": 93,
```
duncanspumpkin, line 8: set the type as `"single_rail_rc"`

duncanspumpkin:
Can you upload the images as well.

objects/official/ride/official.rmct2/object.json
```diff
@@ -0,0 +1,3470 @@
...
+                "VEHICLE_ENTRY_FLAG_RIDERS_SCREAM": true
+            }
+        ],
+        "loadingPositions": [
+            [-2],
+            [0]
+        ]
+    },
+    "images": [
+		{ "path": "images\preview.png", "x": 0, "y": 0 },
```
Gymnasiast, line 95: Should be a forward slash (/).

objects/official/ride/official.rmct2/object.json
```diff
@@ -0,0 +1,3470 @@
+{
+    "id": "official.rmct2",
+    "authors": ["spacek531"],
+    "version": "1.0",
+    "originalId": "00000000|RMCT2   |00003000",
```
Gymnasiast, line 5: This is not correct. It should consist of the DAT header and checksum of the original .DAT object if there is one. If not, let me know and I´ll calculate the values.

objects/official/ride/official.rmct2/object.json
```diff
@@ -0,0 +1,3470 @@
...
+    "objectType": "ride",
+    "properties": {
+        "type": 93,
+        "category": "rollercoaster",
+        "separateRide": true,
```
Gymnasiast, line 10: This property is no longer used, so just leave this out.

Gymnasiast:
I´ll help you get the object working. I have already added some line notes of stuff to address.

Is this supposed to be an OpenRCT2-only object, or do you intend to make it available for RCT2?

duncanspumpkin:
> Is this supposed to be an OpenRCT2-only object, or do you intend to make it available for RCT2?

Its for the new coaster type so it would never load in RCT2.

objects/official/ride/official.rmct2/object.json
spacek531: This is the actual header of the real DAT

spacek531:
I have excluded the images thus far because they are quite large, totalling about 3.5 Mb. I'm told I can compress the images, but I do not know how to do this. If you would like me to upload the images as-is, I can do so.

Gymnasiast:
> Its for the new coaster type so it would never load in RCT2.

@duncanspumpkin It is still possible to have the corresponding DAT file use a different ride type, like X7 did with the RMC trains (which are Twister RC trains in RCT2). Hence the question if @spacek531 was planning to make it available for RCT2 or not.

spacek531:
>Hence the question if @spacek531 was planning to make it available for RCT2 or not.

I am not

Gymnasiast:
OK. Can you zip up the images and drag the ZIP into a comment here? I can take care of compressing them.

spacek531:
[images.zip](https://github.com/OpenRCT2/objects/files/5729283/images.zip)

spacek531:
image

I've tested the object with the new images, and everything works fine. There's a strange freezing going on when hovering over the object in the object selection window, which I will file an issue for in the OpenRCT2 project.

duncanspumpkin (approved):

IntelOrca:
We think using less .png files for the object will help reduce the load times. I.e. storing multiple sprites in the same .png file and referencing their x,y,w,h in the json file. Is it worth doing this before merging?

duncanspumpkin:
Thought we were going to investigate that for later.

IntelOrca:
@duncanspumpkin I know but isn't merging a lot of binary files a point of no return in terms of repository size?

IntelOrca:
@spacek531 I would like to try using one (or few .pngs) containing all the sprites. Are you able to generate a single .png file containing them all and JSON where it specifies `srcX`, `srcY`, `width` and `height` for each sprite?

spacek531:
I will see what I can do

Fred-1044276:
Is there any way you can reduce the train object file size such that it doesn't cause lag/frame rate drops whenever you hover over it in the object selection window?

duncanspumpkin:
@Fred-1044276 that's what we are trying to do and why this hasn't been merged

AaronVanGeffen:
Could it be a solution to embed the images as a packed (RLE-encoded) object, in a manner similar to g2.dat? I realise that defeats the purpose of an easily editable object. However, if the packing is done at release time, that still leaves an easily editable object in the repository.

spacek531:
> @spacek531 I would like to try using one (or few .pngs) containing all the sprites. Are you able to generate a single .png file containing them all and JSON where it specifies `srcX`, `srcY`, `width` and `height` for each sprite?

Is OpenRCT2 currently capable of handling this? I'm looking to finalize this soon.

duncanspumpkin:
I think Ted has added that capability. @IntelOrca could you confirm

IntelOrca:
On NSF, but could be back ported

karst:
I was wondering if it might be an idea to have this (or any of the other two spacek made raptor trains) as additional versions/skins of official object?

oli414:
Superseded by #166

PR #139, opened 2021-01-21
de-DE: Replace ASCII apostrophes

Wuzzy2:
This replaces ASCII apostrophes in the German translation with U+2019.

See also: https://github.com/OpenRCT2/Localisation/issues/2121


Gymnasiast:
Object translations are now also maintained in the Localisation repository. For German, you need to modify this file: https://github.com/OpenRCT2/Localisation/blob/master/objects/de-DE.json

PR #146, opened 2021-05-19
Make RCT1's gray space railings work for everyone & RCT1 queue sign consistency

ZeeMajora:
Makes RCT1's gray space railings work for everyone by using a combination of identical sprites from PATHSPCE.DAT and unused rct1 leftover sprites in g1.dat.
Curiously RCT1 has a different queue sign sprite so i have updated both RCT1 railing objects to use the RCT2 queue sign sprite for consistency with RCT2's railings.
Left = Using RCT1 sprites without RCT1 linked.
Right = Using pathspce + g1 sprites without RCT1 linked.
pathcomparison
RCT2 queue signs on RCT1's railings
queuesigns


Gymnasiast:
That queue sign looks misaligned. It it possible that this is already the case, but I want to check myself before I merge it.

ZeeMajora:
I went ahead and double checked myself and it seems to affect all of rct2's railings aswell.
The layer of their poles also seems to change if the queue path with the sign is a corner piece.
So a fix for this would probably be best kept for a future PR as this seems to be a global issue with all rct2 path railings.

Red queue = RCT1 railings
Yellow queue = RCT2 railings
DeepinScreenshot_openrct2_20210520105816
DeepinScreenshot_openrct2_20210520110600

PR #154, opened 2021-06-05
Fix WW connecting double sided walls

ZeeMajora:
This fixes the maharaja arch pieces & the connecting tudor walls.
As they were not set up as double sided walls so their sprites would change depending on viewing angle.
So i have given them all the isDoubleSided flag and also re-ordered their sprites accordingly.
The maharaja arch pieces were relatively simple to fix and just required referencing the opposite arch's sprites.
The tudor walls were really messed up as they included the wrong sprites so i've completely redone their sprite order to fix that.
For both of them i did have to swap the sprites around to keep the outside sprites the same.
(ex: tmarch1 starts with tmarch2's sprites and vice versa)
Compatibility wise the maharaja gates aren't affected too much, The maharaja palace scenario from wacky worlds looks fine with them after the change, However the inside sprites are different so any parks that used them like that will look incorrect
As for the tudor walls they will technically be different now as they had to be drastically re-ordered but even then the objects themselves were broken to begin with so it's not like builders had much control over them in the first place.

NSF Testing Land 2021-06-05 16-09-24
NSF Testing Land 2021-06-05 16-09-26

### DIFF
--- a/objects/rct2ww/scenery_wall/rct2ww.scenery_wall.tmarch1.json
+++ b/objects/rct2ww/scenery_wall/rct2ww.scenery_wall.tmarch1.json
@@ -6,10 +6,14 @@
     "sourceGame": "rct2ww",
     "objectType": "scenery_wall",
     "properties": {
+        "isDoubleSided": true,
         "height": 8,
         "price": 50
     },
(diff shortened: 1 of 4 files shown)


Gymnasiast:
The preview of all four objects is incorrect:

afbeelding

Would you be able to fix that as well?

PR #156, opened 2021-06-07
Add RCT1/AA/LL music source games.

ZeeMajora:
Should be correct, I simply checked every track on my RCT1, AA, and LL discs to see when new tracks were introduced and added source games accordingly.
The only odd ones are the two custom music tracks as they obviously don't come with the game.
But the CUSTOM.TXT file first appeared in LL so as such i added LL as a source game for the custom music tracks.
I don't have an install of RCT1+AA so i can't test to see if custom tracks were first introduced there but CUSTOM.TXT is not present on my AA disc so i'm doubtful they were introduced in AA.


Gymnasiast:
AA indeed does not seem to support custom music.

PR #159, opened 2021-11-30
Add invisible queue

ZeeMajora:
Adds an invisible queue path surface to complete the invisible path object trio.
NSF Sandbox 2021-11-30 08-51-22


Gymnasiast:
Thanks!

PR #166, opened 2022-03-06
Add hybrid and single rail train

X123M3-256:
This PR adds a vehicle type for the hybrid and single rail coaster.


spacek531:
Are you planning to add the original id to update everyone's rides?
`00000000|RMCT1   |0EFA4E29`
`00000000|RMCT2   |00003000`

X123M3-256:
> Are you planning to add the original id to update everyone's rides?

That's what I wanted to do but it does not seem to work - when I open save files that had the original objects, they still have the original object. I'm not quite sure why this is.

spacek531:
Can you take a picture of how the trains look when updated from older versions?

karst:
Will you create a PR to the main branch for the changelog? Or do you wish me to do it for you, works too.

X123M3-256:
@karst This is a follow up to a previous PR that was merged over a year ago, so all necessary changes to the main branch have already been made, this PR is just for the vehicle objects. See https://github.com/OpenRCT2/OpenRCT2/pull/12110 and https://github.com/OpenRCT2/OpenRCT2/pull/13613/

spacek531:
I believe he is talking about a changelog entry such as "[n] Feature: added hybrid coaster train" "[n] Feature: added single-rail coaster train" because the trains are over a year removed from the rides themselves.

duncanspumpkin:
Changelogs will be handled after the merge when we bump the object revision of the main repo.

karst:
Ahh gotcha!

Gymnasiast:
> OK, I've figured out how to make it replace the old object - turns out that the DAT name in the originalId field must be padded to 8 characters as it is in the DAT file.
> 
> Even with the changes @spacek531 suggests, parks that used the old train will not display properly. The original train was configured differently and I don't see any way to rearrange the cars so that both will draw properly. In addition, the original cars were smaller so even if I could make it so the correct sprites are drawn, they would glitch severely.
> 
> Therefore, I don't see much point it trying to make changes to this object to this end. If this object is to replace the original, doing so will break existing saves, albeit in a minor way. In order to fix the issue, old rides must be double closed and reopened to respawn the train.
> 
> I do not know if this would be considered acceptable, so I'll wait for more feedback before pushing these changes.

Since we have had complaints from NE for far smaller changes, I think it’s probably better not to autoreplace it. 

You should be able to just get rid of the `originalId` parameter.

oli414:
image

I think the restraint color needs a bit of work.

Paiting it gray makes it a very dark gray, I'd expect it to be more in line with the metal bar in front of the seat.
Paiting it white gets a gray that matches with that bar.
As another example painting it light blue makes it seem like its dark blue.

Overall I think the material needs to be a bit brighter to better represent the selected color

oli414 (changes requested):
Just doing this for my previous comment

oli414:
Could you share the 3D model or some pictures of the vehicle in 3D so that we can check if the vehicle is original enough?

SpartanFrederic104:
> image
> 
> I think the restraint color needs a bit of work.
> 
> * Painting it gray (front) makes it a very dark gray, I'd expect it to be more in line with the metal bar in front of the seat.
> * Painting it white (second car) gets a gray that matches with that bar.
> * Painting it light blue (third car) makes it seem like its dark blue.
> 
> Overall I think the material needs to be a bit brighter to better represent the selected color
> 
> Additionally, I think the trains are a bit too close together, those are trains with single cars, and it seems as if they are inside of each other.

Does the hybrid train have any of these issues?

karst:
I wonder if a change of restrain type, say over the shoulder instead of lapbar would be enough of a change to no-longer potentially infringe on copyrights.

X123M3-256:
As far as I'm aware this PR was already rejected, I think they just forgot to close it. There are no RMC hybrid coasters with OTSRs so I'm not going to change that.

oli414:
Alright, yeah since we require some changes that X7 isn't interested in making I'll close this PR. We can create a new PR when an alternative is made.

karst:
I have done an inquiry with RMC and they give a green light to using their trains in OpenRCT2!
image

oli414:
@karst sparked the discussion of asking permission.

We've decided that we will allow recreations if we have permission from the manufacturer or design-holder. However, in the future, permission needs to be given to an OpenRCT2 team member to make sure that we can ask correctly, and archive it if we have been given permission to use a recreation of a product.

karst:
As no model changes are necessary anymore it might be smart to sum up what changes are still necessary for this pr to be approved as currently due to previous situation there's no real overview anymore

oli414:
> As no model changes are necessary anymore it might be smart to sum up what changes are still necessary for this pr to be approved as currently, due to previous situation, there's no real overview anymore

The required changes are already marked as requested changes, you can view them below

SpartanFrederic104:
For those who didn't catch any of the requested changes listed by @oli414 above (these only pertain to the single rail trains):
- The material needs to be reworked so that it better represents the selected color
- The cars could use some extra spacing (not sure how closely spaced the cars are on IRL single rail coaster trains)

karst:
I don't know if it's worth it or not, but is it an idea to give reference to RMC in the object.json, since they have given permission?

X123M3-256:
> I don't know if it's worth it or not, but is it an idea to give reference to RMC in the object.json, since they have given permission?

The object id in the JSON does, but the user facing name doesn't because the vanilla rides don't. I care about consistency more than I care about having the names correct, so unless the vanilla objects are going to be updated I don't really want to change the strings (also, the hybrid and single rail coaster have been in the game under those names for some time now, it might cause confusion if they were changed).

oli414 (commented):
Everything looks good to me.

Could we perhaps change the id name to something more descriptive?

The current name is more in line with the old school file names, but we could just call it "hybrid_coaster" and "single_rail_coaster" or something along these lines.

I guess it might be too late to change that now

oli414 (approved):

karst:
Changing IDs could potentially be annoying for the people who have been using the custom object versions in their parks already. As spacek's raptor trains and these trains have been floating through the community for quite a little now. Ofcourse this is a risk for the people using these objects, but i thought it's worth mentioning

Gymnasiast:
We have code to map old object names in Legacy.cpp. Adding a few entries for the users who used beta versions shouldn’t hurt.

oli414:
Alright, let's change the ID as this is the last chance to do so. @X123M3-256 could you change it to what Gymnasiast suggests above?

> So I would agree renaming
> `openrct2.ride.rmct1` to `openrct2.ride.hybrid_coaster`
> and
> `openrct2.ride.rmct2` to `openrct2.ride.single_rail_coaster`

karst:
May I suggest to name it `openrct2.ride.ibox_train` and `openrct2.ride.raptor_train` or alike so it leaves open space for potentially more trains without things being confusing?

PR #173, opened 2022-04-07
Add alpine coaster vehicle

X123M3-256:
This PR adds a vehicle for the alpine coaster, which is introduced in this other PR https://github.com/OpenRCT2/OpenRCT2/pull/16825. This shouldn't be merged before that one is.

I know there were issues with the performance of the other objects I submitted - this has a lot fewer images, less than the panda train that's already shipped with OpenRCT2, so I'm hoping it will be OK but I'm happy to put the sprites in G1 format if that is preferred.


Dubaifreak:
If The vehicle is approved, i'll use it on a ride!

oli414:
Could you share the pictures of the vehicle in 3D so that we can check if the vehicle is original enough?

oli414:
Currently, this vehicle is not accepted unless we get permission from Wiegand to use their ride vehicle likeness.

oli414:
@Gymnasiast will be contacting Wiegand to ask for permission

SpartanFrederic104:
Just so this object can work with saving track designs, a dummy `originalId` field will be needed like with the single rail, hybrid, and classic mini coaster vehicle objects.

karst:
How much effort would it cost to keep the vehicle looking basically the same in the game (on pixel level) while changing up the model enough for it to not look too similar to Wiegand's design? I should barely need any changes since ADG and Techfun both have really similar designs to Wiegand

Gymnasiast:
Hm, if they have very similar designs to Wiegand, it’s generic enough to merge it (like we did with the track). Do you have some examples of their designs?

oli414:
The model presented here is still very much a recreation of the Wiegand vehicle, there're distinct differences between different manufacturers, so as long as this model can be distinct as well it would be fine to merge it, right now I don't believe that to be the case.



image
image
image

X123M3-256:
> How much effort would it cost to keep the vehicle looking basically the same in the game (on pixel level) while changing up the model enough for it to not look too similar to Wiegand's design? 

It's practically impossible to change the model significantly while making the sprites come out exactly the same, and if I managed to do that the resulting object would be identical to the current one, so I don't think it would change anything.

If I'd made a similar looking but inaccurate model and told them that was the one I used and never shared the real one, perhaps that would have worked - I'm not sure they could really tell from the sprites themselves. But it's too late for that now.

I had actually made a fictional alpine coaster vehicle before I made this one - but that one looks completely different.

karst:
As I have now been told, the copyright issue at hand does not lay with the pixel art. It lays with the model of the vehicle (which is also publicly available) for use for a later on high res graphics pack as well as for when new elements get added. I did do the suggestion it might be an idea that that doesn't become public but only stays in a private archive available for the devs, but I do not know how they feel about that. It would definitely remove the entire issue of the potential infringement.

Gymnasiast:
@oli414 The differences between ADG and Wiegand are _marginal_. Both have a colourful sled with a backrest, a handle to grip, a handbrake, bumpers and lots of other stuff that make them look nearly the same. And if a real-life competitor is able to build such a similar-looking vehicle without any legal trouble, there is no reason why we couldn’t.

oli414:
> Especially since we ran it by Wiegand and they haven’t replied.

If they haven't replied then in my opinion we should assume a no. As per our own policy. Yes the differences are marginal, but also very obvious, if we just change the railing a bit, and the plastic body then we're absolutely fine.

Regardless of the similarities between manufacturers, there're still differences, and we should try to implement these differences in a similar way.

Gymnasiast:
As per our own policy, we do not need to ask permission for generic designs. Under that rule, the Alpine Coaster track type was merged.

oli414:
I don't believe we described that anywhere. 

It still doesn't mean that we can copy a real life design 1:1 (Within the available resources to do so).

X123M3-256:
> As per our own policy, we do not need to ask permission for generic designs. Under that rule, the Alpine Coaster track type was merged.

The alpine coaster track isn't generic, it's also modelled on the Wiegand version (and I said so in the PR for the track). The sprites might not be clearly recognizable due to the low resolution (though I don't think the vehicle really is either), but the model definitely is. The ADG version (which I wasn't aware of when I made this) has different shaped ties and four rails rather than 6.

Gymnasiast:
That was my point: the track was deemed acceptable under the policy, so I don’t see why the same wouldn’t apply for the vehicle. Especially if even actual competitors can make such a similar vehicle.

karst:
Good morning, I have just received an answer on a messages @Gymnasiast and I sent to Wiegand.
Screenshot_20221121-095522

This means they're giving green light! I'll make sure to print the entire email log to pdf later tonight.

oli414:
All clear then, great that they have been able to reply to the request!

Gymnasiast:
This is wonderful news!

@X123M3-256 Is the vehicle ready to merge, as far as you’re concerned?

X123M3-256:
@Gymnasiast Yes

Gymnasiast:
Great! @oli414 can I merge this?

X123M3-256:
Wait, actually I think the spacing might need fixing, should only take a few minutes.

oli414:
Got a notification about the spacing but looks like that wasn't an issue as the comment is now gone. @X123M3-256 could you share some screenshots of the vehicles in a few different colors (Even the bits that are typically black/grey). I'm at work atm so can't check it myself in game at the moment.

Just want to verify that the color representation is consistent like with the single rail coaster vehicle

X123M3-256:
@oli414 

alpine

X123M3-256:
> Just so this object can work with saving track designs, a dummy `originalId` field will be needed like with the single rail, hybrid, and classic mini coaster vehicle objects.

Good catch, I've fixed that.

@Gymnasiast do you want the ID changed to openrct2.ride.alpine_coaster to match the form of hybrid and single rail? If this is done, similar import code would be needed to not break parks that use the current version of the object.

oli414:
Is the seat not recolorable?

Gymnasiast:
@X123M3-256 Yes, please. I’ll make sure to add that code.

X123M3-256:
@Gymnasiast I've updated the ID.

@oli414 The seats are non remappable. I usually find that remappable seats don't look good -  and in this case I can't find a single IRL example where the seat is painted anything but black.

I prefer to use the secondary remap color for a stripe or something like that, but here there isn't really anywhere suitable, so I use just a single remap color. That said, vanilla vehicles almost always have two remap colors, even very small ones. And most do use the secondary remap for the seat color.

I think the unused color slots default to black, so this could potentially be changed without breaking backward compatibility, but so far I've never been happy with the results of making the seats remappable - it usually ends up that you have to pick specific color schemes to get adequate contrast, whereas I want my vehicles to look good with any color choice. This is not the case for the vanilla sprites, but I've never been able to get similar results. And in this case it doesn't seem that the seat is ever colored IRL.

karst:
With the next object release we do need to convert `openrct2.ride.alp1` into `openrct2.ride.alpine_coaster` for all parks that currently use the ride unofficially.

oli414:
Okay that's fine then, I'm happy then, colors look good @Gymnasiast

Gymnasiast (approved):

PR #174, opened 2022-04-12
Fix 1920s Racing Cars getting stuck

ZeeMajora:
The 1920s racing cars from TT can get stuck after a while due to them having a larger spacing value than the vanilla go karts.
To fix this i decreased the spacing value of them from 274912 to 209100 which seems to have stopped them from getting stuck.
I chose this number as numbers around 213000 will cause them to frequently bump into eachother slowing down throughput by a lot and values higher than 223000 will cause them to start getting stuck.
This does have a side effect of making their sprites slightly clip into eachother but i think it's minor enough to make it worth fixing the much bigger issue of them getting stuck.
So a spacing value of 209100 seemed like a good sweet spot that lead to the least sprite clipping whilst also preventing them from getting stuck.
I let them run for 20 ingame years with the spacing value of 209100 and they never got stuck, tests with the vanilla spacing value usually had them get stuck within the first 3 years so this seems to have fixed it.
This will also only affect newly spawned versions of these vehicles, any already existing ones will use the old spacing value until the vehicles are respawned.
Here's some screenshots to showcase the new spacing value compared to the old one.
Old:
DeepinScreenshot_select-area_20220412032335
DeepinScreenshot_select-area_20220412032322
New:

DeepinScreenshot_select-area_20220412034258
DeepinScreenshot_select-area_20220412032220

### DIFF
--- a/objects/rct2tt/ride/rct2tt.ride.1920racr.json
+++ b/objects/rct2tt/ride/rct2tt.ride.1920racr.json
@@ -46,7 +46,7 @@
         ],
         "cars": {
             "rotationFrameMask": 31,
-            "spacing": 274912,
+            "spacing": 209100,
             "mass": 250,
             "numSeats": 1,
             "numSeatRows": 1,


duncanspumpkin:
A change like this would break simulation. Network version must be incremented if this is merged during the next object release

Gymnasiast:
Given how the cars are unusable at the moment, I would like to merge this for v0.4.0 release. Would you be okay with that @duncanspumpkin ? Obviously, I will also bump the network version when updating the object release (I usually do so anyway).

duncanspumpkin:
> Given how the cars are unusable at the moment, I would like to merge this for v0.4.0 release. Would you be okay with that @duncanspumpkin ? Obviously, I will also bump the network version when updating the object release (I usually do so anyway).

yeah its fine, I wasn't sure what you normally do on an object release was just making a comment that it definitely needs to be done for this one.

PR #175, opened 2022-04-14
Add Galaxy Style as object

karst:
I made this PR as an idea. If you guys don't want to add new soundtrack stuff yet that's okay.

The reason I am pushing this as a PR is so we can show the capabilities of the 0.4.0 update. 
This is a custom made song that I made for OpenRCT2.

This song has the following PR on OpenMusic: https://github.com/OpenRCT2/OpenMusic/pull/14

I have pushed this as a WAV file that's rendered out with a bitrate of 22050 and a depth of 8-bit to keep the filesize as low as possible while making it unnoticable in game. I have pushed this as WAV as this currently is the only supported audio format.

Let me know what your guys' opinion is on this idea and if you wish to push it for 0.4.0.

### DIFF
--- a/objects/official/music/galaxy_style/object.json
+++ b/objects/official/music/galaxy_style/object.json
@@ -1,0 +1,22 @@
+{
+    "id": "openrct2.music.galaxy",
+    "sourceGame":"official",
+    "authors": [
+        "Karst van Galen Last"
+    ],
+    "version": "1.0",
+    "objectType": "music",
+    "properties": {
+        "tracks": [
+            {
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Eventually, we certainly want to include new music. But this is a replacement for existing music and we are still working on how to nicely handle that. The v0.4.0 release is also too close to get it in anyway, and we haven’t really reviewed your track either.

Additionally, we will save music objects in the `OpenMusic` repository and not here, even when they eventually end up in an objects release.

PR #193, opened 2022-06-16
Auto package parkobj images

duncanspumpkin:
Still needs the final zipping element and the little bit of workflow adjustment. I would also like to rearrange one or two objects so that their .png files are in an `images` folder.


Gymnasiast:
I noticed two problems:
1. If I download the objects as an artifact from the CI, they are doubly zipped.
2. `rct2dlc.ride.zpanda` does not get an LGX.

IntelOrca:
> If I download the objects as an artifact from the CI, they are doubly zipped.

That is always the case with GitHub upload artifacts. You can ignore this, it won't be the case for releases. The zip inside that zip will be what is on the release.

IntelOrca:
zpanda should now be fixed.
I have also changed it so that any object containing a .png file now gets an LGX for it. So no more .parkobj files contain any .png files, it's either a obj, G1 or CSG reference.

Only downside is that currently `noCsgimages` may have a duplicate or semi-duplicate lgx file, since it is complicated to work out which entries are identical. This stuff will probably change though when asset packs are introduced.

duncanspumpkin:
> zpanda should now be fixed. I have also changed it so that any object containing a .png file now gets an LGX for it. So no more .parkobj files contain any .png files, it's either a obj, G1, CSG or LGX reference.
> 
> Only downside is that currently `noCsgimages` may have a duplicate or semi-duplicate lgx file, since it is complicated to work out which entries are identical. This stuff will probably change though when asset packs are introduced.

I purposely only made it touch the purely png objects as it would be a mess touching the objects that are a mix.

Gymnasiast:
Just tested the objects again (including actually using the resultant objects in OpenRCT2). They seem to work correctly. The objects I tried now load without a noticeable delay, even in debug mode. :+1: 

The objects are still a zip-in-a-zip. I don’t know how difficult it would be to fix this, but if we want to autocreate a release just by tagging, I think we more or less have to.

I don’t have a strong opinion on the mixed objects bit.

IntelOrca:
> I purposely only made it touch the purely png objects as it would be a mess touching the objects that are a mix.

Not sure I understand. It can be messy to implement the script, but I have done that now. The only downside is that you get duplicates with the noCsg arrays, but that will be changed when we use asset packs.

> All we had to do to fix zpanda was to allow for empty strings.

That is all we "had" to do. But I went further so that there was no .png encoded required at all when loading these objects.

Gymnasiast:
I have taken the liberty to merge this, as I would very much like to get an objects release done before I go on holiday. If the commit in question turns out to be a bad idea, reverting it will be simple enough.

PR #195, opened 2022-07-22
Add converted files

karst:
I have converted all these and edited the object.json files to the standard of the other already included files.

I have converted all the files that were listed on the wiki for things that should be featured in the game.

PR #198, opened 2022-07-31
Add pirate roofs

karst:
Seperation from #195 requested by @Gymnasiast 

Includes all pirate roof objects except for the one with a chimney.


Gymnasiast (approved):

Gymnasiast:
That should be all inconsistencies taken out, while remaining compatible with existing saves. They should now behave exactly the same as the two original roofs.

PR #199, opened 2022-08-02
German: path surface/railing, audio, coaster

Wuzzy2:
This adds and updates a couple of German strings. Primarily this affects strings for footpath surfaces, footpath railings, audio and the new special OpenRCT2 rollercoasters.

PS: Is there an automated way to find out which strings of a langauge are still untranslated?

### DIFF
--- a/objects/official/footpath_railings/openrct2.footpath_railings.invisible.json
+++ b/objects/official/footpath_railings/openrct2.footpath_railings.invisible.json
@@ -16,6 +16,7 @@
   "strings": {
     "name": {
       "en-GB": "Invisible Railings",
+      "de-DE": "Unsichtbare Geländer",
       "nl-NL": "Onzichtbare ondersteuningen"
     }
   }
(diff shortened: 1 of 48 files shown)


Gymnasiast:
We now maintain these strings in https://github.com/OpenRCT2/Localisation/tree/master/objects. However, since I _just_ synced the translations a few days ago, I will merge this PR anyway.

I don’t think there is really a way yet to see what hasn’t been translated. We should really write better tooling for that.

Wuzzy2:
Ah, thanks. Good to know. Maybe the objects repository should be archived then, with a link to the correct repo in the description / readme.

tupaschoal:
It is still useful, we track the objects here, but the translations are now drawn from localisation, so that work here can only be about adding new objects and modifying game-related object properties

Wuzzy2:
:D
OK, nvm.

Yes, I saw the new big per-language JSON files in the localization repo. This is *definitely* much better to work with, thank you.

PR #200, opened 2022-08-07
draft of blue grid

733737:
Preview icon doesn't work

https://github.com/OpenRCT2/OpenRCT2/discussions/17723


733737:
I made this because I believe that RGBY makes logical sense instead of RGY Purple; However, I couldn't get the preview icon to work. I want help from others with the translation of "Blue" to other languages and if someone could fix the preview icon implementation for me, I would be very grateful...

Issue #202, opened 2022-08-08
invisible paths still cause random sprites to appear


733737:
The invisible pathing feature still causes random sprites to appear upon loading a save game. My proposal is to replace these objects with a transparent footpath object, instead of a corrupt element there would be a footpath but it would use transparency so it would still be invisible. I am currently working on implementing this

733737:
I have fully implemented this and am waiting on blue land grid to submit PR,. I have tested the new invisible paths and they are working correctly

Gymnasiast:
Can you upload a screenshot of those glitches?

733737:
https://user-images.githubusercontent.com/94884086/183407839-93fbc684-a969-4aec-bcc4-921f2e787af1.png

spacek531:
![image](https://user-images.githubusercontent.com/3454156/187592657-628947a1-5a10-47ee-b742-4e175a1c7fc4.png)
![image](https://user-images.githubusercontent.com/3454156/187592669-e42beae7-9420-413a-843f-c3adfa3bdbcc.png)
Other examples

733737:
![unknown](https://user-images.githubusercontent.com/94884086/192944639-840afb6b-4a57-4d37-999d-0e3cef40d5d9.png)

733737:
![asddsaasddasas](https://user-images.githubusercontent.com/94884086/218676651-d7c00c8f-72a2-4857-8fe6-4f31fdf6c505.png)

duncanspumpkin:
I think this would be better solved in the engine rather than the object.

PR #204, opened 2022-08-08
Replace invisible pathing to fix issue #202

733737:
This will fix the issue where random corrupt elements show up frequently on invisible pathing
Github Park 2022-08-08 03-25-30
Multiplayer Round 59 2022-09-28 13-10-51 https://github.com/OpenRCT2/OpenRCT2/issues/19398


733737:
can someone test whether this will break current parks using invisible pathing?

733737:
It seems the devs are uninterested in my method of fixing this issue.

PR #215, opened 2022-09-11
Draft of RCT Classic Ride and Park Window Graphics

733737:
I have not added code yet to implement graphics on ride window... See https://github.com/OpenRCT2/OpenRCT2/pull/18004


Gymnasiast:
This is something you really have to explain/discuss beforehand (preferably on Discord). Anyway, can you explain what this is, what it does does and why you want to add it?

733737:
I converted the data from the "Projects" section under "RCT Classic Compatibility" that I thought would be most desirable to users, from gif to png and created a dummy object that doesn't have all of the required references. The data is already available on github; I simply want to put it in the game. I personally would use this to change the appearance of the ride open/close/testing and park open/closed buttons in the UI. In order to implement there would need to be a check box added to the theme creator's last tab. RCT1 style lights were added in the past, so I'm sure there is a method for adding UI button replacements. And of course it would be purely aesthetic and totally optional to the user

733737:
I think I will close this as I was mistaken, the elements I found are not ride lights

PR #218, opened 2022-09-15
Clean up multi-dimension coaster train sprites

ZeeMaji:
Removes over 10000 invisible sprites from the multi-dimension train object by making it use new spriteGroups alongside a touched up images table, massively reducing it's sprite count.


Gymnasiast:
I take it you have tested this?

ZeeMajora:
Yeah, it's been a while but i used it pretty frequently when i was testing some custom builds some months ago and never noticed any issues with it.

PR #219, opened 2022-09-15
Make reversed heartline cars use sprites from utcar.dat

ZeeMajora:
This makes the reversed heartline twister cars use sprites from the non-reversed heartline cars dat, as the sprites between the two are identical (including the thumbnail) asides from the vehicles being swapped in the sprite order. This won't make much of a difference to the average user but it does mean that custom installs can omit utcarr.dat to reduce file size (1mb) which i think would be a nice little save in exchange for just a couple of extra lines of image table text in our utcarr json.

### DIFF
--- a/objects/rct2/ride/rct2.ride.utcarr.json
+++ b/objects/rct2/ride/rct2.ride.utcarr.json
@@ -72,7 +72,13 @@
         ],
         "buildMenuPriority": 1
     },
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Putting this on hold. I don’t know whether this a good idea or not yet - we’ll know more once the asset packs turn around.

ZeeMajora:
Closing this as i feel like this would be better suited as a graphics asset pack later on once they are implemented.

PR #220, opened 2022-10-15
Update RMC coasters to use RMC track sound

karst:
Linked to [OpenRCT2#18312](https://github.com/OpenRCT2/OpenRCT2/pull/18312)


karst:
Do the objects also need a version increase? Or should it be fine without it?

Gymnasiast:
Yes, these objects need a version increase. `2.1` will do.

The changes look fine to me, but I guess we’ll have to merge the sound effects and asset packs for them first? (Haven’t yet caught up with those, apart from Karst demonstrating them to me.)

karst:
What could also work is merging this (after i have added the version increase), bumping the object release on here and on that PR. In the end that could also be done afterwards but this way you'd keep it all together.

PR #223, opened 2022-10-29
Add preview images to vanilla music

karst:
I think I am gonna change the preview images to the first game they were introduced in. Need some input on it.

Preview images will look like this: 
image
image


karst:
Why's that failing? Does it not recognize it yet?

IntelOrca:
I am not sure, it seems to fail to import RCT1.png. I assume RCT2.png was fine since it processed candy okay, but fails on dodgems.

karst:
It seems to want to turn the images into gxc which I think is unnecessary for these specific images (since then it'll reuse a lot of sprites). No clue if I can make build ignore the files.

Gymnasiast:
As I said in person, I’d like to provide illustrations that show a setting that fits the music style, rather than album covers or logos. Partly because it looks nicer IMO, but also because I want to avoid suggesting endorsement by using logos.

karst:
yep i agree, preview images like the ones @ZeeMajora made for quite a few rides would be awesome for music

PR #233, opened 2023-01-10
Add portcullis to Walls & Roofs

karst:
Quick PR to add these two to Walls & Roofs. I noticed these weren't part of the sorting by default, while they should be, imo.
image


Gymnasiast:
The reason it’s not part of the group is to avoid always selecting it. Rather, the portcullis objects attach themselves to this group if they are selected.

We _could_ change this, but it will affect more than just sorting. And if we do go on that route, you will need to remove the scenery group association on the objects themselves.

karst:
Ah, that I wasn't sure about. I wasn't sure if the default wall object selection were directly coming from this group or if it was just sorting, as it is a default group. I was hoping it would only affect the sorting in this case but if it doesn't I am not 100% certain about having this done. I personally don't mind but I could see the want to require castle theming selected.

karst:
Do we want this in the game? Otherwise I am closing this.

Gymnasiast:
I’ll put it in the chat, let’s hear what the users think.

SpartanFrederic104:
I'm in favor of adding the portcullis to Walls & Roofs.

PR #235, opened 2023-03-03
fix OpenRCT2#19398 invisible paths still cause random sprites to appear

spacek531:
image

This should fix https://github.com/OpenRCT2/OpenRCT2/issues/19398

By definition a path object must have 52 images. One image is the thumbnail, 51 images are for the 51 possible combinations:
https://github.com/OpenRCT2/OpenRCT2/blob/dd624f9bff4ce116a40f121507a426b484cb9412/src/openrct2/paint/tile_element/Paint.Path.cpp#L58-L67

This object currently has 51 images, not 52. I do not know why this bug does not affect certain parks - it *should* affect all parks.


objects/official/footpath_surface/openrct2.footpath_surface.invisible.json
```diff
@@ -8,7 +8,7 @@
     "noSlopeRailings": true
   },
   "images": [
-    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""
+    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""
```
Broxzier, line 11: The images array basically holds two parts, the preview and the actual sprites. There are 51 sprites for normal paths, this change makes it look like there are 52. No big deal, but I think it would be cleaner to explicitly state that the preview is blank.

```suggestion
    "",
    "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""
```

Broxzier (approved):

PR #236, opened 2023-03-03
Add diagonal sloped bank sprites to hybrid and single rail vehicles

X123M3-256:
This PR adds slopes for diagonal sloped banks to the hybrid and single rail train. It should not be merged before my other PR for large sloped turns because that adds the code needed to load these properly, but it would ideally be merged immediately after that so the trains can display properly on the new pieces.


AaronVanGeffen:
With https://github.com/OpenRCT2/OpenRCT2/pull/19547 having finally been merged, this should PR probably be considered for merging as well.

X123M3-256:
Really should happen soon. I would have made this part of the same PR if the objects weren't in a different repo. The trains will glitch on some of the pieces that were added until this is merged.

duncanspumpkin (approved):

PR #238, opened 2023-03-26
add preview image for invisible queue

733737:
adds preview image for invisible queue


Gymnasiast:
Nice idea.

I’d like to review and test this.

733737:
only minor issue may be that I wasn't sure about the 1 pixel offset and couldn't pixel peep to tell so the image is 1 extra pixel on x and y

Broxzier:
It looks like the feet of the guests are cut off, presumably because this was taken from one of the default paths that cover them? For this path, it makes sense to show the feet properly. Perhaps recreating this setup, then creating a very small "giant" transparent screenshot is the easiest way to make it work.

733737:
I'd prefer if someone else could do this. Personal reasons, right now is not a good time for me to contribute much to OpenRCT, I'm sorry.

733737:
@Broxzier how do you like this original image?

733737:
@Broxzier I redid the image

PR #240, opened 2023-04-22
Fix reversed looping trains friction sound

ZeeMaji:
Was brought up in the discord. Fixes the looping trains having chain lift sounds as their friction sound for their second cars. In vanilla RCT2 this issue couldn't be heard but now with the reversed trains toggle being merged it's now very obvious.


Gymnasiast (approved):

duncanspumpkin:
Addressing this through a different method. Might still be worth while applying this change though.

duncanspumpkin (approved):

PR #241, opened 2023-04-24
Update translations

duncanspumpkin:
need to manually fix some things.


Gymnasiast:
@AaronVanGeffen created a few scripts to do this. These should be documented somewhere.

I was actually planning to do this - let me know if you need help.

duncanspumpkin:
Yeah looks like the scripts break if the hunk contains unrelated changes and the language changes annoyingly.

PR #242, opened 2023-04-30
Update localisation for objects

ZehMatt:
~~I tried my best to keep the diff minimal but the inconsistency is all over the place. I wrote some scripts using NodeJS to also help improve the other repo, not sure yet how the code is best shared among the two.~~

Diff is now quite simple


tupaschoal (approved):
I just find the removals of `en-US` weird

duncanspumpkin:
en-us has been removed if identical to en-gb.

ZehMatt:
> I just find the removals of `en-US` weird

I did that to keep it consistent, sometimes it was present with identical translation to en-gb and sometimes it wasn't, I think due to the way we use en-gb as fallback we can omit them. I can also enforce the other way around and always have en-us there.

PR #243, opened 2023-05-03
Format all json files with new script

ZehMatt:


objects/official/ride/openrct2.ride.alpine_coaster/object.json
```diff
@@ -60,23 +40,16 @@
...
-                ]
+                "loadingPositions": [-1, 1]
             }
         ]
     },
     "strings": {
-        "name": {
-            "en-GB": "Alpine Coaster Cars"
-        },
+        "name": { "en-GB": "Alpine Coaster Cars" },
```
Gymnasiast, line 48: This is going to suck when inserting languages.

Gymnasiast:
I'm not sure this is a good idea. Especially not the bit where I put a line note.

I was actually working on better scripts before exiting the team, I could finish those instead.

objects/official/ride/openrct2.ride.alpine_coaster/object.json
ZehMatt: The languages are synchronized with a new script, the end user doesn't have to deal with this. Also JSON is not made to be human friendly at all, using JSON was the wrong choice so for consistency reasons we decided to format all the objects first with the new script to minimize diffs when synchronized.

ZehMatt: I fixed this.

ZehMatt:
I think this is the best middle ground as for readability and density.

Gymnasiast (approved):
I checked a few files. It’s now much more consistent, good job :+1:


Do you plan to add some kind of CI to ensure that all files stay formatted correctly? (E.g. something like we do with `clang-format` in the main repo?)

ZehMatt:
> I checked a few files. It’s now much more consistent, good job 👍
> 
> Do you plan to add some kind of CI to ensure that all files stay formatted correctly? (E.g. something like we do with `clang-format` in the main repo?)

That's the plan but for now I just want to get this updated so we can do the release. I want to have the scripts somewhere that the localisation repo and this one can use together, its a small collection of js modules that I wrote and since both repos share the same data formats it makes sense to have it on its own repo maybe, I'll have to think a bit about that once I get there.

Broxzier (approved):
This LGTM, I checked ~~all~~ some files to verify no data is lost. I'm all in for consistency.

duncanspumpkin (approved):

PR #245, opened 2023-07-26
Updated sort priorities for expansion scenery groups

### LINKED ISSUE #244: Scenery group ordering
Now that we can have 255 scenery groups visible in the window, it's common for sandbox play to just pick all the groups. But, annoyingly, the expansion groups show up in the middle of the list instead of the end.

image

This ordering is set by the `priority` field of each groups' json file ([example](https://github.com/OpenRCT2/objects/blob/e71cbcbe8ccb87735189abf3255edc75f3a52fbf/objects/rct2/scenery_group/rct2.scenery_group.scgsport.json#L26)). RCT2 groups use priorities 20-39, but for both Wacky Worlds and Time Twister, all groups have priority 25 (the same as Jurassic theming) for some reason (probably Frontier laziness again? who can say). Not only are they in the middle of the group list, but since they all have the same priority, their effective order is inconsistent each time the game is loaded, meaning the player cannot rely on muscle memory to find a particular group.

It would be nice to change the expansion groups' json data to give them priorities above 39, in order to have the groups sorted by game. This is a trivial data change, but I wanted to see if anyone had strong opinions about it.

fidwell:
Updated expansion scenery groups to sort after the base RCT2 groups in the scenery picker, instead of being sorted at random in the middle. The new order is:

- RCT2 groups, as before (1-39)
- Panda theming and any custom groups (40)
- WW alphabetically (41-47)
- TT alphabetically (48-54)
- Six Flags, as before (99)

image

Closes #244.

### DIFF
--- a/objects/rct2tt/scenery_group/rct2tt.scenery_group.scg1920s.json
+++ b/objects/rct2tt/scenery_group/rct2tt.scenery_group.scg1920s.json
@@ -58,7 +58,7 @@
             "rct2tt.scenery_large.hrbwal09",
             "rct2tt.scenery_large.hrbwal11"
         ],
-        "priority": 47
+        "priority": 48
     },
     "images": [
         "$RCT2:OBJDATA/SCG1920S.DAT[0..1]"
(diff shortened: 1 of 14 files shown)


tupaschoal:
It looks ok, I guess. Didn't we need to touch any other sort order?

fidwell:
Did you have something specific in mind? Although there are still a couple RCT2 groups that share a priority (Sports, Wild West, and Pirate are all 38), in order to "fix" that, we'd have to bump _everything_. Since custom groups all use 40, I don't think we should touch anything lower than that.

tupaschoal:
I think that's fine

tupaschoal (approved):

Gymnasiast (approved):

PR #247, opened 2023-09-19
Removing the rct1 tag on 3 items and change 3 other object from rct1 tag to rct1ll tag

TheClaudeQC:
Objects that wasn't in rct1 at all (or uncertain) are:

-The 2 candy "roofs"
-The Small Jurassic Wall

Objects that wasn't in rct1 w/o dlc, but was in Loopy Landscapes are:

-The 2 Circular Castle Towers
-The Lily


TheClaudeQC:
This is now outdated as the small jurassic wall was in added attractions

PR #251, opened 2023-10-08
Rename custom music and mark as compat objects

### LINKED ISSUE #228: Rename CUSTOM1 and CUSTOM2 to "legacy custom music" and flag them as compatibility objects

ZeeMaji:
Renames both of the custom music 1/2 styles to legacy custom music 1/2 and marks them as compatibility objects.

Closes #228

### DIFF
--- a/objects/rct2/music/rct2.music.custom1.json
+++ b/objects/rct2/music/rct2.music.custom1.json
@@ -8,6 +8,7 @@
         "rct2",
         "rct1ll"
     ],
+    "isCompatibilityObject": true,
     "objectType": "music",
     "properties": {
         "originalStyleId": 23,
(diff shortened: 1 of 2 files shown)


Gymnasiast:
I’m pondering whether it would be a good idea to also clear the translations. After all, they are now no longer correct. OTOH, they are not _that_ incorrect.

ZeeMaji:
From what i can remember, previous object PRs that changed the en-gb string for an object never cleared the other translations (eg: the path object renames from a year or so ago) and just left them as is until they were updated in the localisation repo.

To be honest in this case it seems pretty minor as the en-gb name was only slightly changed so the translations are still mostly correct, just slightly oudated.

Gymnasiast:
Yeah, fair enough.

Gymnasiast (approved):

tupaschoal (commented):
Change looks ok, though tbh I don't really know what the impact is

ZeeMaji:
This will make the old custom music objects (the ones that handle vanilla's custom music implementation) compatibility objects, which means they won't be selectable from the object selection window and can only be imported from old parks and deselected. They don't really serve much of a purpose beyond compatibility with old parks since music objects are a far better way of doing custom music now so marking the old custom music objects as compatibility objects seems sensible.

tupaschoal (approved):

Gymnasiast (approved):

PR #257, opened 2023-11-24
Re-introduce RCT1 road

Gymnasiast:
The RCT2 road draws handrails in the tunnels and on slopes (because RCT2 was programmed to do this with all paths), which makes it look weird and also means that roads in RCT1 scenarios don’t always look like they should.

Here are the two side by side (RCT2 on the left, RCT1 on the right):

afbeelding

### DIFF
--- a/objects/rct1/footpath_surface/rct1.footpath_surface.road/object.json
+++ b/objects/rct1/footpath_surface/rct1.footpath_surface.road/object.json
@@ -1,17 +1,18 @@
 {
-    "id": "rct2.footpath_surface.road",
+    "id": "rct1.footpath_surface.road",
     "authors": [
         "Chris Sawyer",
         "Simon Foster"
     ],
     "version": "1.0",
     "sourceGame": [
+        "rct1",
         "rct2"
     ],
     "objectType": "footpath_surface",
-    "isCompatibilityObject": true,
     "properties": {
-        "editorOnly": true
+        "editorOnly": true,
+        "noSlopeRailings": true
     },
     "images": [
         "$RCT2:OBJDATA/ROAD.DAT[71]",


ZeeMaji (approved):

objects/rct1/footpath_surface/rct1.footpath_surface.road/object.json
```diff
@@ -0,0 +1,43 @@
...
+    ],
+    "version": "1.0",
+    "sourceGame": [
+        "rct1",
+        "rct2"
+    ],
+    "objectType": "footpath_surface",
+    "properties": {
+        "editorOnly": true,
+        "noSlopeRailings": true
```
tupaschoal, line 15: Why not just adding this property to the existing rct2 road?

Gymnasiast: That would affect how existing parks look, and that always attracts complaints. So it has basically become policy to make sure all object changes are fully backwards compatible.

The way I did it, both RCT1 and RCT2 road will be drawn correctly, but only one will show up in the Object Selection, so it doesn’t clutter things up.

tupaschoal: Fair enough, I guess

tupaschoal (approved):

objects/rct1/footpath_surface/rct1.footpath_surface.road/object.json
Gymnasiast: Thanks!

ZeeMaji:
Note that in the next objects release, the S4 import table will need to be updated to use the new RCT1 road object when importing S4.

PR #259, opened 2023-12-12
Make RCT1 toilets use noCsgImages

### LINKED ISSUE #253: RCT1 Toilets should use graphics from CSG
Currently the RCT1 toilets object uses graphics supplied from here, these should probably be removed as that's inconsistent with the rest of the RCT1 objects which read their graphics from CSG and have a noCsgImages table for fallback images. And it seems rather dubious to supply unmodified RCT1 sprites in an official object.

ZeeMaji:
This changes the RCT1 toilets to use images from CSG along with adding a noCsgImages table to make it more in line with the rest of the RCT1 ride objects as it predates the noCsgImages system, it will use the RCT2 toilets sprites if the user does not have RCT1 linked and this gets rid of the rather dubious unmodified original game sprites from the object's images folder. This should also slightly reduce the file size of the object.

Note that this will not affect the appearance of the RCT1 toilets for users with RCT1 linked, only those without RCT1.

Screenshot_select-area_20231211215143

Closes #253


Gymnasiast (changes requested):
Like I said, I am against this. The only thing this will achieve is degrade the experience for people who do not have RCT1 linked.

ZeeMaji:
I mean it technically is a downgrade yes, but to be fair this _is_ an RCT1 object so i don't really see why people without RCT1 should be prioritized. I doubt there's very many people who would use the RCT1 toilets without having RCT1 linked given how most people who want RCT1 content would want the full experience, rather than just the scraps of the like two RCT1 objects that work without RCT1 linked.

I really don't see why the RCT1 toilets need special treatment here, sure they were added before the noCsgImages system was introduced and have better artwork but i still don't really see that as a compelling reason for openrct2 to continue to supply the unmodified original sprites when we have a new system that allows for using the original sprites directly from RCT1 and gets rid of the potential head ache of dubiously supplying the original commercial game's sprites for this.

Gymnasiast:
> [It] gets rid of the potential head ache of dubiously supplying the original commercial game's sprites for this.

No, it does not. Far from it, in fact. It might have done if g2.dat, or the DLC content did not exist.

And this is the only benefit you mentioned - that against the drawback of degrading the experience for a considerable group of players makes it absolutely not worth it IMO.

ZeeMaji:
Fine, if you're so adamant about keeping the images pre-packed as they are i don't think i can change your mind.

PR #263, opened 2024-01-15
Remove improper voffset flags

ZeeMaji:
Most small scenery objects in WW/TT seem to have this flag, but these three scenery pieces weren't designed with it in mind so their sprites are improperly offset upwards which would cause their sprites to shift positions when rotating your view. Removing the flag fixes this.

Oil barrel on green tile is for reference, it's not changed
Before:
Screenshot_select-area_20240115093546

After:
Screenshot_select-area_20240115093456


Gymnasiast (approved):

tupaschoal (approved):
Nice find!

PR #270, opened 2024-01-17
Split dam tower & brick base block objects

ZeeMaji:
Splits these two objects into two objects each, the original objects are marked as compat objects while the new ones are not. The required metadata changes for id, scenery groups, and originalid have been made so the new objects are the ones that are used in new parks.

For the original dam tower, the support rendering broke a park so the original object has it's supports removed, while the new object retains support rendering.
For the base block, the original object would block supports entirely which was used by a few parks as deliberate support blockers. While the new object allows supports to connect to them (like the vanilla base blocks)


Gymnasiast:
I ran the images through `sha1` and saw that all of them are identical. Could you change it so it only has a single image in both the old and the new object?

ZeeMaji:
Done, although it will still take 4 image slots ingame due to small scenery needing 4 images and i believe the object builder will still duplicate the images in the lgx file it builds. resulting in no file size change.

PR #272, opened 2024-01-25
Add RCT1 ice cream stall

ZeeMaji:
Adds the RCT1 ice cream stall as a compat object. Due to the object's RCT2 counterpart having non-remappable sprites in it's dat i've supplied copies of the remappable sprites we use for the RCT2 ice cream stall to be used for the fallback images. I've changed their file names to nocsg_# to make this more clear in the images folder that these are solely for fallbacks.

preview


objects/rct1/ride/rct1.ride.fruity_ices_stall/object.json
```diff
@@ -0,0 +1,77 @@
...
+        "",
+        "",
+        { "path": "images/nocsg_0.png", "x": -23, "y": -44 },
+        { "path": "images/nocsg_1.png", "x": -23, "y": -44 },
+        { "path": "images/nocsg_2.png", "x": -28, "y": -46 },
+        { "path": "images/nocsg_3.png", "x": -28, "y": -39 }
+    ],
+    "strings": {
+        "name": {
+            "en-GB": "Fruity Ices Stall",
```
Gymnasiast, line 37: RCT1 referred to this as the "Ice cream stall", RCT2 as "Fruity Ices Stall", possibly in an attempt to differentiate it from the Ice Cream Cone Stall.

Should we go with RCT1 description, or keep RCT2’s? What do you reckon?

ZeeMaji: Imo i'd just keep the name and descriptions the same as RCT2 where applicable. RCT1 has a lot of weird naming differences that'd just be confusing if you were to place them in an RCT2 context (eg: The lay-down coaster is known as the flying coaster in RCT1) especially since the RCT2 ride type names and descriptions are still used. Plus it would probably be easier for translation purposes since the same translations can be used for both RCT1 & RCT2 objects in most cases while if different strings were used those would have to be separately translated.

tupaschoal (approved):

PR #280, opened 2024-01-27
Add all RCT1LL coaster trains

ZeeMaji:
Adds all RCT1LL coaster trains as compat objects. Also adjusts the buildMenuPriority of the RCT1 2-across invert trains and the RCT2 4-across invert trains to fix some oddities with which icon is displayed when grouping by ride types. Now the values are as follows:

RCT2 Vekoma Giant Boomerang Trains: 1
RCT1/2 Vekoma Face-off Trains: 2
RCT2 Vekoma SLC Trains/RCT1 Giovanola Invert Trains: 3
RCT1/2 B&M 4-Across Trains: 4


Gymnasiast:
Could you explain why it is necessary to also modify the RCT2 trains?

ZeeMaji:
It's mainly due to how the 2-across trains are available for both the compact + b&m inverted coasters, as the 2-across trains use the same buildMenuPriority for both ride types.

With the old 2-across value of just 1, they would get overwritten by the face-off trains on the compact invert, and by increasing their value to 3 that would make them overwrite the 4-across trains on the b&m invert. So as such the RCT1LL & RCT2 4-across trains have had their priority set to 4 so that way they will always overwrite the 2-across trains while still ensuring that the 2-across trains will always overwrite the face-off cars (and giant boomerang trains if you enable them with cheats or in custom scenarios.)

The only RCT2 train modified by this is the 4-across trains, i just posted the others to give more context as to what the buildmenu priorities for all of them would be in relation to one another.

Gymnasiast (approved):

tupaschoal (approved):

PR #281, opened 2024-01-28
Add most RCT1AA coaster trains

ZeeMaji:
Adds most RCT1AA coaster trains, the only ones that are missing are the classic wooden twister trains & lay-down coaster trains as OpenRCT2 cannot support them yet. A couple of things to note:

1: The virginia reel tubs have the same doubled sprites as mentioned for the rapids boats in the water rides PR, so everything about that applies to the reel tubs too.
2: I've went ahead and included the reversed heartline cars. this feels mildly redundant as we're phasing out reversed vehicle objects but the current import code relies on actual objects right now from what i can tell. If you'd rather me not include them if you'd rather just change the import code to import the non-reversed cars as reversed or something similar to that then let me know and i'll remove them from this.


Gymnasiast:
I’m going to look into the reversed cars issue.

Gymnasiast:
Can you take the reversed Heartline Twister train out for now? I might need a bit more time to properly look into it and I want to merge the rest of it.

ZeeMaji:
Done.

Gymnasiast (approved):

tupaschoal (approved):

PR #294, opened 2024-02-18
Wooden support structure additions

ZeeMaji:
Adds three new wooden support structure scenery pieces to the walls and roofs group. They are:

A half height wooden support structure small scenery piece.
A wooden support structure wall piece.
And a half height wooden support structure wall piece.

Screenshot_select-area_20240217211803
Screenshot_select-area_20240217211828


Gymnasiast:
Aren’t (at least some of) in g1.dat? I’d prefer referencing that to copying them if possible.

ZeeMaji:
The half height support structure sprites unfortunately could not be read from g1, as their offsets aren't correct when used as small scenery even while removing the voffset flag. The images could be removed in the future once we do have original asset offset overriding support in objects.

As for the walls, i don't believe they exist in g1.dat as i checked all of the sections of g1 that have support sprites and didn't see standalone wall sprites for either height.

deurklink:
Do you think a triangular piece could be added as well? I usually copy them from a wooden coaster track piece now: 
image

ZeeMaji:
You could technically add them, but they probably wouldn't work well as scenery supports currently only allow for square shaped supports which wouldn't look good with these diagonal support structures, and if you forgoe the lower supports then you would have to place stacks of them and they would also not connect to sloped/underwater terrain well which i don't think would be up to the official additions standard.

deurklink:
> You could technically add them, but they probably wouldn't work well as scenery supports currently only allow for square shaped supports which wouldn't look good with these diagonal support structures, and if you forgoe the lower supports then you would have to place stacks of them and they would also not connect to sloped/underwater terrain well which i don't think would be up to the official additions standard.

That explanation makes sense. I'll just keep copying them then!

Gymnasiast (approved):

tupaschoal (approved):

PR #295, opened 2024-02-20
Add flipped version of post scenery wall

ZeeMaji:
Adds a flipped version of the post scenery wall as an official scenery addition, this doesn't require any new sprites as they can just be read from the original object's dat.

Screenshot_select-area_20240219203247

Curiously, scenery walls reserve image slots for sloped images even when they're not enabled by the object. this is why there is 4 empty images in the images table.

### DIFF
--- a/objects/official/scenery_wall/official.scenery_wall.post_flipped.json
+++ b/objects/official/scenery_wall/official.scenery_wall.post_flipped.json
@@ -1,0 +1,44 @@
+{
+    "id": "official.scenery_wall.post_flipped",
+    "authors": [
+        "OpenRCT2 developers",
+        "Simon Foster"
+    ],
+    "version": "1.0",
+    "sourceGame": "official",
+    "objectType": "scenery_wall",
+    "properties": {
+        "isBanner": true,
+        "hasPrimaryColour": true,
+        "height": 4,
+        "price": 70
+    },
+    "images": [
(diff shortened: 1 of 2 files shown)


objects/official/scenery_wall/official.scenery_wall.post_flipped.json
```diff
@@ -0,0 +1,44 @@
...
+    "objectType": "scenery_wall",
+    "properties": {
+        "isBanner": true,
+        "hasPrimaryColour": true,
+        "height": 4,
+        "price": 70
+    },
+    "images": [
+        "$RCT2:OBJDATA/WALLPOST.DAT[6..7]",
+        "",
```
Gymnasiast, line 18: Can you explain these empty images?

ZeeMaji: Scenery walls seem to reserve image slots for slope sprites even if they're not enabled by the wall object. You can see this in the vanilla post's images when you export them.

I recall you mentioned that scenery walls will still display slope images even with the flag not being set, so i assume this is how that works.

Exported vanilla post images, there's no empty images after 7 so the empty images only exist for the first set of sprites (0 thru 5).
Screenshot_select-area_20240220050849

Gymnasiast (approved):

tupaschoal (approved):

PR #304, opened 2024-03-07
Add sloped sprites to wallbb8, wallbb16, and wallbb32

fidwell:

Thanks to ZeeMaji for advice.

wallbb8
wallbb16
wallbb32


ZeeMaji:
Have you tested to ensure they line up correctly with the RCT1 brick terrain edge?

fidwell:
Yeah thanks for reminding me. Hard to tell if they're _perfect_ (they should be) but they're at least pretty dang close:

image

ZeeMaji:
They seem to be off by one pixel vertically from some testing i did, i think they need to be shifted up by one pixel?

fidwell:
For all 3 walls, there is a gap if I shift images 2 and 3 up. But for images 4 and 5 it looked better.

ZeeMaji:
By shift up, i meant the texture of the wall in relation to the full wall sprite so you would just raise the full wall sprite by one pixel before you apply a sloping mask or something. not shifting the sprite's offsets up or down. Kind of hard to explain but i hope that explains it well enough.

For what it's worth, i think it looks mostly fine as it is right now. Just being nit picky as it'd be preferable to get this as good as possible in the initial PR so as to not have to make more changes later on.

fidwell:
I think I get what you mean. I tried shifting up the texture one pixel, but I think it was better before; in this version, it looks like the brick on the seam is too tall.

image

ZeeMaji:
Hmm, odd. Well then in that case i suppose just go with what you had before.

Issue #157, opened 2021-11-25
Adding slopes for wall scenery


dr-dirt:
Most, if not all, vanilla wall objects are only able to be placed on flat land.

This has some limitations to what they can all be used for (ie. as fences on sloped tiles, lining the roofs without sides).  Implementing this wouldn't necessarily be difficult as editing the flat walls to create sloped sprites isn't too difficult:

<img width="422" alt="Unnamed park 2021-10-10 20-36-01" src="https://user-images.githubusercontent.com/57810480/143379122-430e6ebc-96d6-4d89-a171-bc06ee1012a3.png">
(The above includes some walls not in the base game)

Alternatively, perhaps this task could be automated.  The .parkobj objects from the original objects seem like they use the sprites from the original DATs so I'm not sure how exactly this would look adding additional images to those.

fidwell:
Since I've been working on this lately, I figured this ticket is a good place to log progress. Here's a list of the walls I think make good candidates for having sloped sprites.

Not complete:
- Matte wall (wallbr8, wallbr16, wallbr32)
- Matte beige wall (wallcb8, wallcb16, wallcb32)
- Shiny wall (wallcf8, wallcf16, wallcf32)
- Rusty wall (wallrs8, wallrs16, wallrs32)
- Corrugated steel wall (wallsk16, wallsk32)
- Poles (walltn32)
- Wooden wall (wallwd8, wallwd16, wallwd32)
- Brick wall (wbr1a)
- Stone wall (wbr2a)

In PR:
- Brick wall (wallbb8, wallbb16, wallbb32) #304 
- Brick wall (wbr1a) #302 

Completed:
- Railing (wallmm16, wallmm17) #291 
- Fence railing (wallrh32) #303

ZeeMaji:
I think all of the ones you have planned would fit well in the game.
Of the ones under consideration, the three steel walls would probably work well especially as i seem them used as general walls a lot due to their texture & diagonal pieces so having slopes for them seems pretty desireable. The igloo & roman walls i'm mildly unsure about but they would probably still be sensible.

fidwell:
Sounds good to me, thanks for the input!

fidwell:
With that last PR complete, I've finished all the wall types I wanted to do for now. I think that more than satisfies the original issue, and we can close it out.

I've looked briefly at the Roman and corrugated steel walls and don't feel great about the results I've had so far. Maybe I (or someone else) will come back to them some day.

[...]

PR #306, opened 2024-03-08
 Add sloped sprites to walltn32

fidwell:

Pinging @ZeeMaji for review. This was an easy one though

walltn32


ZeeMaji:
The sprites themselves look fine to me, but i'm wondering about the usecases for this. specifically non-hacked usecases. I don't think that's too big of a deal for this but it would be nice to know what usecases there are for this outside of hacked parks since official additions (i see these new sloped sprites as another form of official additions) should preferably compliment all aspects of the game.

fidwell:
I hadn't really thought much about it... tbh I don't use this object at all myself, so maybe there is no good use case. I won't be upset if you think this doesn't really fit in and we don't include it.

I'm keeping track of all of these in this ticket: #157 Maybe it'd be good to go through it and see if you have concerns about any of the unfinished walls before I get to them.

ZeeMaji:
In that case i don't really feel like this fits well for official inclusion currently, it doesn't really seem to compliment anything that comes with the game.

As for the others, i'll give my thoughts on them in #157 proper.

Gymnasiast:
I don’t think it would hurt to include these, to be honest, especially considering we’re just adding additional frames to an existing object. I can see it being useful for queue lines, just like some of the others.

ZeeMaji:
When you put it that way, i suppose it would be fine to include.

Gymnasiast (approved):

tupaschoal (approved):

PR #307, opened 2024-03-09
 Add sloped sprites to wallbr8, wallbr16, and wallbr32

fidwell:

Pinging @ZeeMaji for review.

wallbr8
wallbr16
wallbr32

Comparison with RCT1 terrain edge (for texture; the color can't be matched):

image


ZeeMaji:
Hmm right i forgot about that, the colors not matching is kind of unfortunate but i don't think it impacts the walls on their own too much.

On their own, i think they look fine. Nice work.

ZeeMaji (approved):

tupaschoal (approved):

PR #311, opened 2024-03-10
Add sloped sprites to wallcf8, wallcf16, and wallcf32

fidwell:

wallcf8
wallcf16
wallcf32

One interesting thing about this wall is that (I think) it is meant to match the shiny castle walls, which have sloped sprites but those sprites do not line up with the ground—they are shifted up by a pixel, so they hover off the ground. In this screenshot, on the left are two 16-high walls, and on the right is a 16-high wall with a castle wall atop it. You can see the seam. This seems like an issue with the castle wall itself and I don't think that wallcf should replicate this behavior.

image


ZeeMaji:
How do they look when they're combined with the gray sandstone edge?

fidwell:
image

ZeeMaji:
Looks good to me.

ZeeMaji (approved):

PR #312, opened 2024-03-10
Add sloped sprites to wallwd8, wallwd16, and wallwd32

fidwell:

wallwd8
wallwd16
wallwd32

Comparison with RCT1 terrain edge (for texture; color can't be matched). This one is slightly different as the diagonal crossbeam textures don't exist on sloped terrain, only for the full-face texture. Not sure if that is worth trying to replicate here.

image


ZeeMaji:
Probably fine to forgoe matching it with the terrain edge if the textures & colors are different.

ZeeMaji:
Looks good to me.

ZeeMaji (approved):

PR #317, opened 2024-03-11
Add sloped sprites to wallig16

fidwell:

wallig16


ZeeMaji:
The vertical offsets seem a bit off here, they look like they need to be shifted down?

fidwell:
You're right; shifting them down a pixel looks way better.
image

ZeeMaji:
Looks good to me now.

ZeeMaji (approved):

PR #319, opened 2024-03-14
Fix RCT1 corkscrew train's slopes42 group

ZeeMaji:
Fixes an issue with the RCT1 corkscrew coaster using an incorrect sprite in it's slopes42 group due to the last sprite of the upward portion of the group having a typo. As i had to use the original unformated images table to do these changes this changes the format of the images table a bit. I hope this isn't too big of a deal as it should just get fixed during the next localisation sync?


Gymnasiast:
I’ll run the formatter over the branch, that should make the diff a little bit cleaner. (We should should make it easily available and use it on the CI, but I’d have to discuss that with Matt.)

Gymnasiast (approved):

tupaschoal (approved):

PR #320, opened 2024-03-18
Fix mandarin duck boats capacity

ZeeMaji:
This vehicle erroneously has it's seats set to 4 when only 2 guests are visible on the boat at a time. Due to some weirdness with how the swan boats animation style (animation: 2) is handled i've switched them over to the water trikes animation (animation: 5) which lets us shave off some dummy blank images the original object had to include as a hacky workaround for the swans animation style always expecting 2 sets of guest sprites (+1 for the animation if you want to count that) in a strange order.

On old saves that still have the duck vehicles spawned, they will retain their old capacity of 4 so this doesn't affect old saves. However due to the change in animation style there's a chance that on the first initial tick the sprites will be glitched as i believe it's still technically using the swans animation style for the first tick which will result in the second animation frame using some images from outside of the object's images table. After the first tick it fixes itself as it switches to the trikes animation so it's not super noticeable but i felt like it was worth mentioning just incase.

For the capacity string i've only changed en-GB, not really sure if i should change the others as it'd probably be cleaner to change them on the localisation repo.

### DIFF
--- a/objects/rct2ww/ride/rct2ww.ride.mandarin.json
+++ b/objects/rct2ww/ride/rct2ww.ride.mandarin.json
@@ -112,22 +112,22 @@
         },
         "capacity": {
             "en-GB": "2 passengers per boat",
-            "fr-FR": "4 passagers par bateau",
-            "de-DE": "4 Fahrgäste pro Boot",
-            "es-ES": "4 pasajeros por bote",
-            "it-IT": "4 passeggeri per imbarcazione",
-            "nl-NL": "4 passagiers per boot",
-            "sv-SE": "4 passagerare per båt",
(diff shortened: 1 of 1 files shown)


Gymnasiast (approved):
Looks alright to me. Could you update the other languages anyway, though?

ZeeMaji:
Hmm, are you sure it would be a good idea? There's some object string changes in the localisation repo so i'm not sure if the next localisation sync would overwrite this.

ZeeMaji:
Hmm, well alright.

Gymnasiast (approved):

tupaschoal (approved):

PR #325, opened 2024-04-05
Fix y offset of sloped castle walls

Gymnasiast:
Good Knight Park before:
Good Knight Park 2024-04-04 21-00-32
and after:
Good Knight Park 2024-04-04 20-59-40

Sloped castle wall on top of sloped sandstone wall, before:
afbeelding
and after:
afbeelding

Crazy Castle perimeter wall, before:
image
and after:
image

### DIFF
--- a/build.mjs
+++ b/build.mjs
@@ -86,12 +86,12 @@ function isImageLgxCompatible(image) {
             return false;
         }
     } else {
-        return true;
+        return image.hasOwnProperty('path');
     }
 }
 
 function isImageLgxRequired(image) {
-    return typeof image !== 'string';
+    return typeof image !== 'string' && image.hasOwnProperty('path');
 }
 
 function shouldProcessImageArray(images) {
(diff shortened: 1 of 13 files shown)


ZehMatt:
Just to make sure, does it render like that in the original too?

Gymnasiast:
Here is Good Knight Park in vanilla RCT:

image

ZeeMaji:
For some very strange reason, the objects builder is lumping all of the wall objects into a single parkobj in the rct2 objects folder as  "rct2.scenery_wall.wallbadm.parkobj"

ZehMatt:
> Here is Good Knight Park in vanilla RCT:
> 
> image

They seem to be fine here? Where is the difference coming from, thats the same asset right?

build.mjs
```diff
@@ -86,12 +86,16 @@ function isImageLgxCompatible(image) {
             return false;
         }
     } else {
-        return true;
+        if (image.hasOwnProperty('path')) {
+            return true;
+        } else {
+            return false;
+        }
```
AaronVanGeffen, line 93: We could do without the branch here, right?
```suggestion
        return image.hasOwnProperty('path'));
```

Gymnasiast: I guess.

ZeeMaji:
Objects seem to be building fine now, and as for the fix itself the gaps between the sloped walls and the floor are gone and look much better now. Nicely done.

ZeeMaji (approved):

PR #333, opened 2024-05-10
Fix wall settings on WW/TT large scenery

Gymnasiast:


ZeeMaji:
I tested these large scenery items (at least two of each of the sizes listed in the original issue for this) thus far and with the exception of the luxury liner, they all appear to be working fine.

Screenshot_select-area_20240511023908


The luxury liner pieces technically don't have anything wrong with them but due to the way they're intended to be used you can put walls between them which loods odd. So they should probably block walls on the edges that are meant to connect to other parts of the ship.

Screenshot_select-area_20240511023130

Gymnasiast:
Thank you for testing! Agreed on the ocean liner. Will get on that when I’m home and have some time.

PR #342, opened 2024-07-22
add rct1 "scrambled eggs" preview colour

733737:
adds the rct1 colour preset scheme to the ride presets
Twist_RCT1_Icon

### DIFF
--- a/objects/rct2/ride/rct2.ride.twist1.json
+++ b/objects/rct2/ride/rct2.ride.twist1.json
@@ -27,6 +27,13 @@
             ],
             [
                 [
+                    "white",
+                    "light_blue",
+                    "black"
+                ]
+            ],
+            [
+                [
                     "bright_red",
                     "white",
                     "black"


objects/rct2/ride/rct2.ride.twist1.json
```diff
@@ -24,6 +24,13 @@
                     "yellow",
                     "black"
                 ]
+            ],
+		    [
```
ZeeMaji, line 28: This extra bracket should be removed, i don't think the object will load correctly with this present.

733737: the way it is now follows the convention of the object. Removing it isn't logical as far as I know.

Gymnasiast: Looks like it is correct, it’s just that the indentation is off.

ZeeMaji: My bad, i guess it is just the indentation.

733737: it appears correct in notepad++, which is what I used to edit it. I suppose I could edit the indentation in the browser.

733737:
I haven't looked into it. I was watching twitch the other day and someone was streaming rct deluxe and I noticed the twist thumbnail. I could simply google all of the rct1 thumbnails from the wiki and compare them to the ingame presets however.

Gymnasiast:
It looks like you mixed tabs with spaces (even after the latest commit). Please use spaces throughout.

733737:
should be fixed

PR #343, opened 2024-07-23
Include RCT1 Ferris Wheel Preview Colour Preset

733737:
add Ride Color Scheme Preset from RCT1 Thumbnail to Ferris Wheel
Ferris_Wheel_RCT1_Icon


ZeeMaji (approved):
It's a pleasing, unique color scheme so i think it works nicely as a default preset.

PR #344, opened 2024-08-02
Add wall version of beta brick terrain edge

### LINKED ISSUE #340: Scenery wall versions of beta brick edge
I feel like it would probably make sense to include scenery wall versions of the beta brick edge, since if it had been carried over to the final version of RCT1 it likely would have been turned into scenery walls in RCT2. So this would make it more consistent with the other "building" terrain edges and also make the texture more flexible to work with for builders.

fidwell:
Resolves #340. Created a new wall object that match the texture of the brick wall beta terrain edge.

image


ZeeMaji:
I'm a bit unsure if it would make sense to include the rock walls as there's no wall version of the final/RCT2 rock edge.

For the names i would just shorten them to "Brick Wall" and "Rock Wall" or even just "Wall" as the wall names in vanilla are pretty generic.

For the object ids should remain "official" as they were not a part of the rct1 beta, they're just official additions based on the beta assets.

fidwell:
I didn't even consider that some of the terrain edges don't have corresponding walls. I probably wouldn't have made the rock wall if I'd thought about it.

If anyone feels strongly about _not_ having them, that's fine by me.

tbarford:
Maybe add that wall texture as an object, too?

Gymnasiast:
I generally want to keep the brakes on adding too much extra content. Could you put forth your arguments for including these as official content?

fidwell:
Having the beta brick wall would create a nice parity with how the existing brick terrain edge has a wall that goes with it.

The rock wall doesn't fit as well and I don't have any strong arguments about keeping that one.

karst:
I could see the brick wall fit in with the default objects, I could see rock wall (as well as for the non-beta versions) be a non-essential official object. 
@Gymnasiast good ideas for the object store, too?

ZeeMaji:
For what it's worth, the brick wall was the only one i mentioned on the issue that prompted this PR. #340.

fidwell:
I've rebased this PR and limited the change to just adding the brick wall.

tupaschoal:
@karst @ZeeMaji @Gymnasiast ?

karst:
I'm happy with this. 
I do want @Gymnasiast's opinion though. He has ideas on which objects to release with the main package and which objects would be additional content.

ZeeMaji:
Ingame the sprites look good to me. However i would change the walls & roofs scenery group so they're listed between the final brick walls & iron walls as their current placement at the very end of the group seems strange.

fidwell:
I reordered the group and added "OpenRCT2 developers" to the credit list. That feels more appropriate to me, since neither of us made new art; we just converted pre-existing assets.

Gymnasiast:
> I'm happy with this. I do want @Gymnasiast's opinion though. He has ideas on which objects to release with the main package and which objects would be additional content.

I think this would be more suitable for a featured/add-on set, rather than directly bundled with the game.

PR #347, opened 2024-09-13
Add object with original game peep names

AaronVanGeffen:


tupaschoal (commented):
It would be very nice if we could have names in other languages eventually. Do you think it is easily doable with the changes that start from this?

AaronVanGeffen:
Hmm, I honestly think we shouldn't try to 'translate' names at an object level. Personally, I'd like to have, say, a set of Spanish names as a separate object to tie them to particular scenarios, regardless of what interface language is being used.

tupaschoal:
> Hmm, I honestly think we shouldn't try to 'translate' names at an object level. Personally, I'd like to have, say, a set of Spanish names as a separate object to tie them to particular scenarios, regardless of what interface language is being used. 

My point was not that we should try to translate, but whether we would be able to easily plug another language name set in, and make it more "immersive" for people not playing in English.

objects/rct2/peep_names/rct2.peep_names.original.json
```diff
@@ -0,0 +1,1056 @@
+{
+    "id": "rct2.peep_names.original",
+    "authors": ["Chris Sawyer"],
+    "version": "1.0",
+    "sourceGame": ["rct2"],
+    "objectType": "peep_names",
+    "strings": {
+        "name": {
+            "en-GB": "Original game peep names"
```
Gymnasiast, line 9:
```suggestion
            "en-GB": "Original peep names",
            "nl-NL": "Originele bezoekersnamen"
```

groenroos:
> Hmm, I honestly think we shouldn't try to 'translate' names at an object level. Personally, I'd like to have, say, a set of Spanish names as a separate object to tie them to particular scenarios, regardless of what interface language is being used.

I'd suggest having a bank of names for each language, and adding an option for the player (right after the language dropdown?) to choose whether they'd prefer names that always match their chosen UI language, or names that 'fit' the scenario they're playing.

Not all scenarios (esp. RCT1) are unambiguously set in a specific geographic location though, so I wonder what would be the best thing there.

PR #349, opened 2024-09-13
Update CI to use non-deprecated upload-artifact action

AaronVanGeffen:


tupaschoal (approved):
Might be worth scouring all our repos to update them all hahaha.

And then keep track of whenever we update it somewhere, to update elsewhere.

PR #352, opened 2024-10-31
Sort RCT1 Scenario text objects

ZeeMaji:
This reorganizes all of the RCT1AA & LL scenario texts to use the proper rct1aa & rct1ll object ids & source games, rather than just lumping everything under rct1.

This also moves fort anachronism to the official category with an id prefix of rct1dlc as it was released as a DLC scenario for RCT1 ala panda world for RCT2, it's not included in all copies of LL afaik so imo it doesn't really fit well there. If the other RCT1 DLC scenarios (magazine promo parks, competition parks) get scenario text objects in the future then they should also be classified as official with an id prefix of rct1dlc.


Gymnasiast:
>  it's not included in all copies of LL afaik

I have yet to come across one that doesn’t include it, to be perfectly honest.

ZeeMaji:
I'm swear my old disc copy of LL did not include it as i have no memory of the park from before i got RCT1 Deluxe. Either way it's source game is still official in my opinion if we're classifying panda world as official. As both parks originated as DLC but were packed in with later releases. (Panda world is packed in with RCTC)

Gymnasiast:
Alton Towers also originated as DLC, but we classified it as LL nonetheless.

I agree that RCTC muddies the waters, as we classified "Tycoon Park" as rct2.

Maybe I should ask it like this: how would you classify Panda World, Tycoon Park, Alton Towers and Fort Anachronism in a consistent way?

ZeeMaji:
For Alton Towers i do feel like it's LL source game is stronger than the other DLC parks, as unlike them Alton Towers received a lot of updates that take advantage of the new features available in AA & LL (Features like 4-across invert trains, an actual proper enterprise ride, path railings, and some other things) while the other DLC parks were unchanged so the version packed with LL is quite different from the original DLC version.

For Panda World & Fort Anachronism, as well as the magazine promo parks & competition parks should those get stex objects in the future i feel like official makes the most sense as they don't originate in any particular pack as they originated as DLC, and when they were packed in with the full packs they were unchanged from their DLC origins. If you were to change them to their base game then that would be inconsistent with other dlc objects in the repo right now which are categorized under official. (eg: panda trains for junior coaster)

For Tycoon Park, i didn't realize it was here but now that you've mentioned it i'm a bit unsure. Currently the only other RCTC object uses RCT2 as a source game (rct2.audio.base.rctc) so Tycoon Park is technically consistent with that for what that's worth. It's kind of a weird case because it's not really DLC as it's solely available via RCTC as base game content so it doesn't really feel like something apart of official to me. But we don't have a dedicated RCTC source game either and this is basically the only thing it would be useful for so i can't imagine it would be worthwhile to add. So i feel like it's mostly fine leaving it's source game as RCT2 for now.

PR #354, opened 2024-11-09
Sort language dump output by object ids

AaronVanGeffen:


language_dump.py
```diff
@@ -114,6 +113,9 @@ def dump_translation(language_to_extract, args, dump_file_name):
     translation_progress = round(100 * translated_str_count / reference_str_count, 2)
     print(f'{language_to_extract}: {translation_progress}% completeness')
 
+    # Sort output by object ids (dict keys)
+    strings_by_object = dict(sorted(strings_by_object.items()))
```
Gymnasiast, line 117: Have you checked that this leaves the keys within an object in the correct order?

AaronVanGeffen: Yes. This was used to generate the dumps in my other PR.

PR #356, opened 2024-11-19
Add RCT1AA wooden twister coaster trains

ZeeMaji:
Adds the RCT1AA wooden twister trains for the classic wooden twister roller coaster.

I'm unsure on how to add co-authors but if spacek could be added as a co-author for this that would be great, as he did a lot of the work on this one in particular from what i recall.

preview


ZeeMaji:
Fallback images need to be fixed.
Screenshot at 2024-11-21 11-41-14

PR #357, opened 2024-11-21
Add initial versions of peep animation objects

AaronVanGeffen:
This is a supporting PR to https://github.com/OpenRCT2/OpenRCT2/pull/23328. The objects proposed in this PR were dumped using a [Python script](https://gist.github.com/AaronVanGeffen/46e41c1aba71cd41bc0446e0c6542e8c).

To do:
- [x] Order string table by language/key
- [x] Correct image table to account for rotations
- [x] Add inline image sprite ids
- [x] Add legacy positions to animation groups
- [x] Remove legacy `entertainerCostumes` array from scenery groups
- [x] Entertainers: drop unused `eatFood` animation
- [x] Deduplicate guest and security animations (inherit from first group?)
- [x] Format animation sequence in one line?? -> not supported by our formatter
- [x] Add RCT1 source games where appropriate

I'd prefer this PR to be squash merged into one commit.


ZeeMaji:
I noticed that the RCT1 source games were missing from the objects when objects shared between both RCT1 & RCT2 define RCT2 as the primary source game with whatever RCT1 pack they originated in as a second source game. From some research i did these should be the RCT1 source games for all of the peep animation objects. 

RCT1

Guest
Handyman
Mechanic
Security Guard
Entertainer Elephant
Entertainer Panda
Entertainer Tiger

RCT1AA
Entertainer Roman
Entertainer Gorilla

RCT1LL

Entertainer Astronaut
Entertainer Knight
Entertainer Snowman

Anything not listed here is RCT2 only (eg: pirate costume).
These should be correct from the research i did, the RCT1 base game only has the base three costumes when i checked the RCT1 demo in a scenario with the roman theming available, the roman & gorilla entertainers are after the standard three in https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/rct1/Tables.cpp with the gorilla costume appearing in the RCT1AA DLC scenario PC Player, and the latter ones after it are all from scenery groups that originated in RCT1LL

AaronVanGeffen:
Good idea. I'll amend my script and redump.

ZeeMaji:
They're in the wrong order as the first entry is for the primary source game, it should be like this:
```
    "sourceGame": [
        "rct2",
        "rct1ll"
    ],
```

AaronVanGeffen:
@ZeeMaji Thanks, I had no idea. I've changed the order to what you suggested.

Gymnasiast (approved):
I haven’t extensively checked the objects, but since this already got reviewed by Maji, I think we can proceed.

One request: we generally have the `en-GB` string on top in the translations. Could you change that order? If not, no big deal.

PR #359, opened 2024-12-17
fix water drawing in hovercar preview image

733737:
fixes issue with object where the water color would not reflect the currently loaded palette (partial fix for https://github.com/OpenRCT2/objects/issues/217 )
before:
Hover_Car_Ride_RCT2_Icon

after:

Final Image approved by @ZeeMaji . Allegedly the original image has incorrect terrain at the top, so it cannot be replicated.


ZeeMaji (approved):

ZeeMaji:
Image file name should probably be changed to preview.png for consistency with other parkobjs with thumbnails, but otherwise looks good to me.

Gymnasiast (approved):

Gymnasiast:
> Allegedly the original image has incorrect terrain at the top, so it cannot be replicated.

To me, those tiles just look like flat tiles, with the tiles in front of them raised (so that you’d see cliff edges if you rotated the view). But I think the smoother terrain in your version looks better anyway.

ZeeMaji:
The full explaination i have is the top of the original thumbnail seems like it was editted in from an earlier take of the thumbnail on accident, as the way it blends with the sloped tile is unnatural. It all looks like it's completely flat but then there's a slope that connects two different heights which just looks odd, i tried multiple combinations of terrains to try and recreate it but what's there in the original is seemingly editted.

PR #375, opened 2025-01-31
Add initial versions of climate objects

AaronVanGeffen:
These are intended to replace the hardcoded climates in Climate.cpp. Weather patterns are defined for each of the eight in-game months.

This opens up the possibility to ship with new climates as well, e.g. climates that feature snow and blizzards instead of rain and thunder.

Dumped using a [Python script](https://gist.github.com/AaronVanGeffen/e05dd9c2943207d563d76fc866848e94)


ZeeMaji:
According to https://rct.wiki/wiki/Climate, the RCT1 source games for these are all from the RCT1 base game except for the cold climate which originates in RCT1LL.

xrgtmx:
Hopefully this feature request of mine can be a topic of discussion: https://github.com/OpenRCT2/OpenRCT2/discussions/23753

tupaschoal:
So far so good, it seems! :)

Gymnasiast:
> Hopefully this feature request of mine can be a topic of discussion: [OpenRCT2/OpenRCT2#23753](https://github.com/OpenRCT2/OpenRCT2/discussions/23753)

I suggest we limit the scope to the original 4 to begin with.

AaronVanGeffen:
> In the most recent commit the RCT1 source game for the cold climate got changed to RCT1 from RCT1LL, was that intentional?

Oops. No, thanks for catching that. I'll put it back.

objects/rct2/climate/rct2.climate.cool_and_wet.json
```diff
@@ -0,0 +1,145 @@
...
+    "authors": [
+        "Chris Sawyer"
+    ],
+    "version": "1.0",
+    "sourceGame": [
+        "rct2",
+        "rct1"
+    ],
+    "objectType": "climate",
+    "scriptName": "coldAndWet",
```
Gymnasiast, line 12:
```suggestion
    "scriptName": "coolAndWet",
```
(this is what’s currently passed onto scripts, I checked)

objects/rct2/climate/rct2.climate.cool_and_wet.json
```diff
@@ -0,0 +1,145 @@
...
+    "strings": {
+        "name": {
+            "es-ES": "Frío y húmedo",
+            "nl-NL": "Koel en nat",
+            "pl-PL": "Zimno i mokro",
+            "en-GB": "Cool and wet",
+            "it-IT": "Freddo e umido",
+            "vi-VN": "Cool and wet",
```
Gymnasiast, line 20: Duplicated from English.

objects/rct2/climate/rct2.climate.cool_and_wet.json
AaronVanGeffen: It's also the name of the object, so it's likely down to me adding the `scriptName` property manually, and _of course_ introducing a typo in the process 😄

ZeeMaji:
Some minor things i noticed, but firstly each object has an empty images table that should be removed as you told me these don't do anything for climate objects.
Secondly en-gb should probably be moved to the top of the strings list for each object since presently it's just mixed with the other strings and i recall this was done for the peep animation objects.

PR #376, opened 2025-02-19
Skidoo > Snowmobile renames

ZeeMaji:
The term skidoo seems to be trademarked and primarily applies to a specific brand of snowmobiles rather than being a generic term for them. (https://ski-doo.brp.com/us/en/legal-notice.html) And also this like it would violate the no advertising clause or whatever it was that got some of the easter eggs and the overt six flags branding removed.

So i feel like these strings should be changed just to be on the safe side.

I have also updated en-US here, if that would be better suited in the localisation repo i can remove it from here and update it there instead. Other language strings have not been updated and some of them do seem to refer to them as skidoo so those will need to be changed too.

### DIFF
--- a/objects/rct2ww/ride/rct2ww.ride.skidoo.json
+++ b/objects/rct2ww/ride/rct2ww.ride.skidoo.json
@@ -53,8 +53,8 @@
     ],
     "strings": {
         "name": {
-            "en-GB": "Skidoo Dodgems",
-            "en-US": "Skidoo Bumper Cars",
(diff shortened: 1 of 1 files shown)


Gymnasiast:
It’s easiest if you just update en-GB and create an issue on the Localisation repo.

ZeeMaji:
I'll create the issue after this is merged.

Gymnasiast (approved):

PR #378, opened 2025-02-26
Add new color presets to panda trains

matheusvb3:
This adds these three new presets to the panda trains:

panda3

Umber and white for brown panda, all umber for grizzly bear and all white for polar bear

### DIFF
--- a/objects/official/ride/rct2dlc.ride.zpanda/object.json
+++ b/objects/official/ride/rct2dlc.ride.zpanda/object.json
@@ -30,6 +30,21 @@
                     "black",
                     "white",
                     "black"
+                ],
+                [
+                    "umber",
+                    "white",
+                    "black"
+                ],
+                [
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Honestly, I don’t think they work. They all look off to me.

matheusvb3:
Those combinations were suggested by Magi and I think they look fine, I had tried one based on the red panda and that looked really off. I also tried one based on pink and blue teddy bears and those looked nice, but it would be a stretch to add those I think. In any case I think it would be good to have at least one alternative to the default black and white, I think it's one of very few trains that don't have any alternatives by default

Gymnasiast:
> I think it's one of very few trains that don't have any alternatives by default

That’s hardly surprising, considering how bad it looks if you select anything other than black and white.

For that reason, I’m going to close this.

PR #379, opened 2025-02-28
Add RCT1 color-like presets to Mine Train RC

orbifant:
Brown RCT1 color scheme for Mine Train Coaster:

openrct_mine2

Not identical but similar to the RCT1 green color scheme for Mine Train Coaster:

openrct_mine1

### DIFF
--- a/objects/rct2/ride/rct2.ride.amt1.json
+++ b/objects/rct2/ride/rct2.ride.amt1.json
@@ -31,6 +31,20 @@
                     "black",
                     "bordeaux_red"
                 ]
+            ],
+            [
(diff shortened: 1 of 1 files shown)


ZeeMaji:
I do like this idea, i pondered about doing it myself at some point.

However, i would like to see the dark green & black preset have it's third color set to black so it matches the original preset better.

orbifant:
image

like this? It does look better, I agree

ZeeMaji:
Looks good to me.

ZeeMaji (approved):

PR #380, opened 2025-02-28
Add RCT1 color preset to Junior RC log trains

orbifant:
Adding RCT1 dark brown/green color scheme to Junior RC log trains

openrct_junior

### DIFF
--- a/objects/rct2/ride/rct2.ride.zlog.json
+++ b/objects/rct2/ride/rct2.ride.zlog.json
@@ -38,6 +38,13 @@
                     "bordeaux_red",
                     "black"
                 ]
+            ],            
+            [
+                [
+                    "dark_brown",
+                    "dark_green",
+                    "black"
+                ]
             ]
         ],
         "cars": [


ZeeMaji:
This seems a bit too similar to the dark brown + black preset already in the game i feel.

orbifant:
It does and I personally like this one better than brown+black, thats part of why I wanted to add it since I assume, deleting existing ones is not an option

ZeeMaji:
I don't think we should remove existing vanilla RCT2 presets.

To be honest for this vehicle specifically i feel like it's pretty different enough from it's RCT1 counterpart that adding the RCT1 presets to it doesn't make sense.

For one it's for a completely different ride type (junior coaster vs classic mini coaster) and the vehicle designs differ. Especially for the remapping as the RCT1 log trains are a lot darker so the same color presets don't really translate over to the RCT2 log trains exactly either.

Gymnasiast:
In that case, I am going to err on the side of caution and close the PR.

PR #381, opened 2025-03-04
Fix OpenRCT2/OpenRCT2#7222: Transparent pixels in sloped path tunnels

mixiate:
This fixes OpenRCT2/OpenRCT2#7222 transparent pixels in sloped path tunnels.

I've extended the sprites of these path tunnels with parts of their sloped square tunnels, to fill the gaps. Depending on the path you could see through them, and they cause the hall of mirrors glitch if nothing is behind. Especially noticeable if paths are set to transparent.

tunnelglitch2 tunnelfix2

tunnelglitch
tunnelfix


Gymnasiast:
@ZeeMaji Could you test this one?

ZeeMaji:
Sure.

mixiate:
Yeah I had thought about this. I haven't tried it in the code yet but when I first looked at it I thought it wouldn't be possible to draw the existing sloped tunnel sprite behind this one because it was too big, but it might actually not be. I'd have to try it. It would add extra complexity to the tunnel painting code though and considering some of the other changes i've been wanting to make, i'm trying to avoid doing more things like that.

If it works i'd be happy to do it that way though. Let me know what you guys think.

mixiate:
Here's what it looks like if you draw the sloped sprite behind. I thought maybe the bounding boxes would make this complicated but they're already lowered, so this sprite that is slightly larger at the bottom doesn't glitch. The stucco tunnels are a little bit awkward. It should of course work with any existing land edges though.

tunnelalternative

mixiate:
I haven't fixed this yet but there is another issue with tunnels that will definitely require a sprite edit as the front sprites simply include too much of the background sprites:
 

tunnelissue

It's not really a reason to not fix this issue in the code, I guess I thought i'd mention it as there will still be a need to edit some sprites and existing custom content will still have that problem unfortunately.

ZeeMaji:
Yeah in that case i think the fix here for OpenRCT2/OpenRCT2#7222 (sprite edits) will be the way to go for all of these issues. It is unfortunately going to persist on existing custom content but this seems like the best way to go about it.

I'll look into reviewing this PR soonish.

ZeeMaji (approved):
Fixes the initial issue and all of the parks with path tunnels i looked at still render as expected or they're fixed now.

Gymnasiast:
@mixiate Could you rebase?

mixiate:
I've rebased.

PR #383, opened 2025-04-16
Disable base colour on non-remappable WWTT vehicles, change black to light_blue

ZeeMaji:
Disables the base colour on all non-remappable WWTT vehicles, i've also change all instances of black in their color presets to light_blue as primary remap seems to have been an off blue color going by the evo of RCT so this is more accurate. This fix was already used for the hippo submarines so this just applies it to all objects.

I also properly disabled remaps on the dhow boats & huskie vehicles as i had missed them during the previous WWTT preset fixes.

### DIFF
--- a/objects/rct2tt/ride/rct2tt.ride.bmvoctps.json
+++ b/objects/rct2tt/ride/rct2tt.ride.bmvoctps.json
@@ -19,7 +19,7 @@
         "carColours": [
             [
                 [
-                    "black",
+                    "light_blue",
                     "bright_pink",
                     "yellow"
                 ]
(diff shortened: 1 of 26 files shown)


Gymnasiast:
Has this led to noticeable differences in how these objects look?

ZeeMaji:
For this PR specifically from what i can tell, no. The only object that had a really bad primary remap were the hippo submarines that were already fixed a while ago. This just simply applied that fix to all non-remappable WWTT objects for safety just incase any rogue primary remap pixels were on any of the other vehicles. Similar to what we did for the second & third remaps in previous PRs.

PR #386, opened 2025-05-08
Add minimap and screenshot images to scenario text objects

AaronVanGeffen:
This add minimap and screenshot images to the scenario text objects, and renames the object type to scenario meta.

Needs additional work to get these loaded into the game and indeed the scenario selection window. I'll work on that longer term, just wanted to file this for now.


AaronVanGeffen:
There appears to be a solid white line at the bottom of all dumped minimap images. This is a bit surprising; I don't know the cause yet. Dump code @ https://github.com/AaronVanGeffen/OpenRCT2/commit/726096181e5318c8b21ddab43c638ce83a0a2d08

objects/official/scenario_meta/rct1dlc.scenario_meta.fort_anachronism/images/screenshot.png
Gymnasiast, line 1: Showing the back of the entrance doesn’t make for a very good preview, methinks. Also applies to some of the other screenshots, like that of Panda World.

PR #389, opened 2025-06-19
Define clearance height of flat rides

J-PIE-314:
This is based on how tall the sprites look in-game. (Does not address any sprite clipping that will occur.)

### DIFF
--- a/objects/rct1/ride/rct1.ride.dodgems/object.json
+++ b/objects/rct1/ride/rct1.ride.dodgems/object.json
@@ -12,6 +12,7 @@
     "properties": {
         "type": "dodgems",
         "category": "gentle",
+        "clearance": 48
         "carsPerFlatRide": 12,
         "carColours": [
             [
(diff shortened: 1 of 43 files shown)


J-PIE-314:
This park shows the heights I chose for each ride and which ones will have clipping issues: [Ride Clearance Park.zip](https://github.com/user-attachments/files/16971725/Ride.Clearance.Park.zip)

objects/rct1/ride/rct1.ride.dodgems/object.json
```diff
@@ -12,6 +12,7 @@
     "properties": {
         "type": "dodgems",
         "category": "gentle",
+        "clearance": 48
```
Gymnasiast, line 15:
```suggestion
        "clearance": 48,
```

J-PIE-314:
I do not observe any clipping of the dodgems and flying saucers with or without your change. I do observe clipping with other rides such as the hall of mirrors: 
screenshot of Hall of Mirrors clipping

Gymnasiast:
@J-PIE-314 What does that do on develop? I expect this to have the same problem there.

J-PIE-314:
> Could you try that?

I looked into this, I could use your assistance in this regard.

PR #392, opened 2025-07-06
Change roller coaster trains to use scream set #2

Gymnasiast:
This is to be merged in conjunction with an upcoming PR to the main repo.


ZeeMaji:
I don't disagree with the way this is being handled, but is there no better way to fix this on the side of the main repo? As changing it like this would leave all of the legacy custom content vehicles that exist and use scream set 0 unaccounted for.

PR #393, opened 2025-07-21
Add Taiwan Park scenario_meta

ZeeMaji:
Adds the scenario meta file for Taiwan Park, the preview images were done by Aaron so if they could be added as a co-author that would be nice.

For the authors i've went with "Infogrames Taiwan Limited" which seems to be correct going by this wikipedia article https://zh.wikipedia.org/wiki/英寶格

For the id/source game i've went with 'rct1tw' and official as that just seemed to make the most sense as this is a pretty odd case.


AaronVanGeffen:
If we're going to customise the preview, what would be preferable? A more zoomed-out overview of the entire island?

ZeeMaji:
I feel like a more zoomed out overview would stick out too much, a custom one will already stick out a bit due to it being inconsistent with the rest of the official scenario previews.

Perhaps just the area after the entrance with the rows of palm trees would work? It's sort of an entrance area in a way.
ksnip_20250722-043348 webp

AaronVanGeffen:
The new screenshot looks good to me. I haven't tested yet, but do the water effects affect the screenshot properly as well?

AaronVanGeffen:
> Although it seems to animate slightly differently from other previews, i'm not sure if it has to do with the exact tick of the water palette animation i took the screenshot on or what.

Yes, it's probably that. Doesn't sound like a problem in my book, but I'll check it out later just to double check if it stands out.

AaronVanGeffen (approved):

PR #398, opened 2025-08-09
Bring in cs-CZ updates

tupaschoal:
@tygrysek90 does this sound right?


tygrysek90 (commented):
I find this perfect (I checked randomly about 20 translations and they are good)

Please merge
Thank you very much

PR #401, opened 2025-08-16
Fix down sloped round tunnels glitching

mixiate:
This fixes down sloped round tunnels glitching as trains go through. For some reason parts of the background sprites are in the foreground sprites, this cuts and pastes those sections in to the background sprites. There's not really any other way to fix this other than editing the sprites, so any custom content will still have the issue unfortunately. This fixes it for all RCT1, RCT2 and ORCT2 edges.

For example, here are the original sprites:
   

And this fix:
    

This is similar to the other tunnel sprites, only these down slope ones have this issue.

https://github.com/user-attachments/assets/73112f06-3503-4192-8f29-9f21a05e0d8d

https://github.com/user-attachments/assets/9853bc83-3920-4690-9aea-d96fe09f171f

There's still a few pixels glitching on some edge and track combinations, I did the best I could, some tracks aren't really fitting in to the tunnels in the first place and there has to be at least a little bit of the bottom of the tunnel in the front sprite or it doesn't look right for the majority of tracks.


tupaschoal:
@ZeeMaji would you mind testing these?

ZeeMaji:
Sure.

PR #406, opened 2025-09-04
Convert track designer grid to object

ZeeMaji:
Converts the dark blue full tile grid surface from the track designer into it's own proper terrain surface object with support for slopes & underground images.

With this, now you can use this surface in your parks and this will allow for enabling land tools in the track designer. Along with getting rid of a pretty nasty paint hack Aaron told me about as apparently the track designer uses the grass surface with a paint hack to render as this blue grid instead.

Since the original texture originates in RCT2, i've marked the source game as that. As for the authors i've just left CS & SF as i'm unsure who could have made the original texture due to it's simplicity, and i've added 'OpenRCT2 developers' since this is a pretty big extension as the surface has went from 1 image, to 37.

I have also made the sprites recolorable similar to the LL space grid surfaces, this isn't usable in the actual game as they are hardcoded to dark_blue. But this does future proof them for whenever we get recolorable terrain surfaces. Having them set to dark_blue matches the original grid's colors perfectly.

For the name, i've opted for 'Designer Grid' as 'Grid' is already taken by the LL space grid surfaces.

Due to how thin the outline is as it's only one pixel thick on the original flat sprite, i had to do some tricks to get it to be scaled properly on the other slope angles. So some terrain configurations with edges might not have an outline. Although this already happened with the original flat sprite on the edges of the map so hopefully it isn't too big of a deal? I can't see any way around this other than completely redesigning the original flat sprite but i'd rather not replace the original design.

ksnip_20250904-074539 webp
ksnip_20250904-074543 webp


Gymnasiast:
Merged into https://github.com/OpenRCT2/objects/pull/447

PR #411, opened 2025-09-30
Fix errors in unnatural water palettes

ZeeMaji:
This fixes two errors in the orange & acid & pink water palettes where they were modifying colors beyond the water colors. 

Firstly, pink tones 2-6 are changed to be the same as natural water on a new separate object. The pink colors in the original orange/acid water palettes seem to have been accidentally leftover from an earlier development version of the game's palette as they appear in some evo of RCT screenshots. For the sake of visual backwards compatibility i've split the objects here since this impacts the looks of old parks with these palettes. The old object has no changes to the pink tones but are marked as compat objects, the new objects have the error fixed.

Secondly, the third to last color (a weird unused blue color) has been changed to the same value it is in the natural & green water palettes. This color seems to be entirely unused which shouldn't affect any parks visually so i've unified it across all the objects, even the now compat marked ones.


ZeeMaji:
Marking as draft as pink water is also affected.
ksnip_20251001-121344
ksnip_20251001-121346

Gymnasiast:
I would prefer also leaving the weird blue as-is in the old objects. It doesn’t make a lot of sense to fix stuff in an object that is left in for compatibility, and I’m not 100% sure it is unused in the first place.

Another question I had was: what does RCT1 do if you set it to orange water? Does it use the same "broken" palette as RCT2 does, or the fixed one?

And finally: do you think we should have a special palette for RCT1 base game imports, considering it used different shades of yellow?

ZeeMaji:
> I would prefer also leaving the weird blue as-is in the old objects. It doesn’t make a lot of sense to fix stuff in an object that is left in for compatibility, and I’m not 100% sure it is unused in the first place.

They aren't referenced anywhere in the code from what I remember when I researched this a while back, the values correspond to the first three remap slots afaik and seemed like weird leftovers that ended up in the palette objects somehow.

> Another question I had was: what does RCT1 do if you set it to orange water? Does it use the same "broken" palette as RCT2 does, or the fixed one?

Very interestingly, the pink palette change seems to be specific to RCT2, as both RCT1 palettes use the same pink tones that differ from what RCT2 has on it's natural water palette. I will need to adjust this PR for this find as we'll need an RCT1 natural water palette compat object to account for this.

RCT1:
rct1-natural
rct1-orange

OpenRCT2:
openrct2-natural
openrct2-orange

> And finally: do you think we should have a special palette for RCT1 base game imports, considering it used different shades of yellow?

This is not possible to do, the palette is the same. It's just the way the dark yellow color is mapped is different in the RCT1 base game, I checked this myself because I was going to make an object for that years ago but it's just not a thing that can be replicated with current palette objects since it's particular to the mapping of the dark yellow color.

ZeeMaji:
Alright, this should be ready for review again. I've undid the change to the unused blue on the old water objects and implemented a compat object for RCT1's natural water palette.

I did investigate further and found no other palette differences between RCT1 & RCT2 beyond an odd case with the # 374b4b color as it seems to have been introduced in RCT2 as it's slot in RCT1's palette corresponds to a seemingly unused # ff00ff placeholder.

Gymnasiast:
Could you explain the addition of the RCT1 natural water? (You may have explained it earlier, but I may have forgotten and I can’t tell from the conversation on this PR.)

ZeeMaji:
The different pink shades were also in RCT1 on the natural water, so I added an object for that as well. This whole thing with the pink shades changing between objects seems to have been introduced from an update in RCT2 during it's development to the natural water palette that wasn't applied to the orange water based palette objects.

RCT1 releases - has the original pink shades.
RCT1LL releases, introduces orange water palette, still has the original pink shades.
RCT2 releases, updates the pink shades but only updates them on natural water and the green water which was based on natural water, forgot to update orange water so it and all of the objects that descend from it use the original RCT1 pink shades.

This PR fixes all of that by creating new objects that have the updated RCT2 pink shades on the unnatural water palette objects (orange water, acid green water, pink water), and also adds an RCT1 natural water compat object that has the original RCT1 pink shades. Here's some images and captions to describe the changes a bit better.

Original pink shades - RCT1 natural water, RCT1/RCT2 orange water, RCT2 acid green/pink water.
ksnip_20260401-223143

New pink shades - RCT2 natural/green water. Fixed orange/acid/pink water objects from this PR.
ksnip_20260401-223151

AaronVanGeffen:
> For the sake of visual backwards compatibility i've split the objects here since this impacts the looks of old parks with these palettes.

If it's not too much trouble, could you give an example of how such a park might be adversely affected?

Gymnasiast:
> > For the sake of visual backwards compatibility i've split the objects here since this impacts the looks of old parks with these palettes.
> 
> If it's not too much trouble, could you give an example of how such a park might be adversely affected?

See the screenshots above. You can see there is a difference in darkness.

PR #414, opened 2025-10-15
Add campaign objects

kyphii:
Implementation for campaigns / scenario packs in object format. These are all the campaign tabs from the current hardcoded implementation. "Other Parks" has been split into Official DLC, Competition Maps, and UCES pack, as the Other Parks will be utilized in game as a tab for loaded scenarios which are not grouped into a campaign. 

Object properties were lifted from my proof of concept non-object implementation, and have been validated in-game via that POC. Translations have been copied over from the relevant StringIDs in the main repo translation files. JSON validity has been confirmed for all files.

One open question is how to handle category / difficulty headers. In my current implementation, these are derived from the scenario category values and displayed dynamically, same as the previous implementation. If we want the headers to instead be controlled by the campaign file, that data will need to be added here.


ZeeMaji:
The source games in the objects seem to have a couple of errors.
1: There are invalid source games set in the former extras tab campaigns, they should have their source game set to ``"official"``
2: The source games set on the LL, WW, & TT campaigns are invalid, they just simply need the underscores removed from their respective entries.
3: The 'Real' Parks campaign should list RCT1LL as a secondary source game, as LL is where this tab was introduced.

ZeeMaji:
Also could we rename the "Original DLC Parks" tab to "Extras"? I think that's a bit more of a generally applicable name for it, as most of the parks contained within it are not DLC. Also "Extras" is more official as it's what the tab that contained these scenarios was called in RCT1.

kyphii:
> I wonder if it will even be possible to use custom headers for campaigns besides Beginner, Challenging, Expert, etc.

Yes, this is supported.

PR #423, opened 2026-01-29
Fix OpenRCT2/OpenRCT2#25720: TT Jet Plane 2 has holes

mixiate:
The sprites for this object are split so weirdly and there's lots of overlapping and bits getting cut off, in the melee this little gap has opened up.
This fills in the holes by the minimum amount necessary by copying this square of pixels over to the sprites on that tile.
Untitled


Gymnasiast:
This has actually CAUSED a hole to pop up for me: 

image

(Notice that the capacity string of the Blob ride is fixed, which shows that I am using the latests objects release)

PR #434, opened 2026-04-20
Remove supports from pirate ship

ZeeMaji:
Removes support rendering from the pirate ship large scenery object with the usual new object + mark old as compat approach to ensure old parks still render as expected.
Left: New
Right: Old
ksnip_20260420-125843


ZeeMaji:
On hold until large scenery blocking supports flags are added at the request of spacek.

PR #435, opened 2026-04-26
pl-PL: Fix misconverted regional characters

mataha:
Excluding mapping `pl-PL` to cp1250 in objexport (though I doubt that makes *any* difference, given Polish strings were extracted using a different method), this is a mirror of OpenRCT2/Localisation#3459.

### DIFF
--- a/objects/rct2/ride/rct2.ride.submar.json
+++ b/objects/rct2/ride/rct2.ride.submar.json
@@ -120,7 +120,7 @@
             "pt-BR": "5 passageiros por submarino",
             "cs-CZ": "5 míst v každé ponorce",
             "ja-JP": "各鑑5名",
-            "pl-PL": "5 pasażerów na łódţ podwodną",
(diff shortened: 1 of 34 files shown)


janisozaur (approved):

Gymnasiast:
The translations are synced from Localisation, so that PR was enough. 

As for the encoding: the Polish edition does not use Windows-1250 - it should have, but it uses a mangled version of ISO-8859-1 for all languages.

Gymnasiast:
As mentioned on Discord, the encoding scheme used is not Windows-1250 but a modified ISO-8859-1 (Windows-1252 is derived from this), also used by all other languages. Support may need to be added still, but it’s much more than just an entry in a table. And slot 12 may have been reserved for Polish but isn’t actually used in the Polish version.

As the translations are synced, everything we can use from this PR is now present, so I’ll close it.

PR #436, opened 2026-05-03
Replace “Kremlin” reference with “St. Basil’s Cathedral”

Gymnasiast:
All pieces seem to be inspired by St. Basil’s Cathedral, which is next to the Kremlin, but not part of it. (Together, they count as one World Heritage site.)

For reference, this is that cathedral:

afbeelding

And this is the Kremlin:

afbeelding

I would like confirmation from @andolga that I got this correct.

### DIFF
--- a/objects/rct2ww/scenery_large/rct2ww.scenery_large.rkreml08.json
+++ b/objects/rct2ww/scenery_large/rct2ww.scenery_large.rkreml08.json
@@ -43,7 +43,7 @@
     ],
     "strings": {
         "name": {
-            "en-GB": "Kremlin Large Tower Base",
+            "en-GB": "St. Basil’s Cathedral Large Tower Base",
             "fr-FR": "Base de grande tour du Kremlin",
             "de-DE": "Kreml - Großer Turm - Basis",
             "es-ES": "Base de torre grande del Kremlin",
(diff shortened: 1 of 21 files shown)


andOlga:
Yes, this cathedral exists outside the kremlin walls (and always has -- in fact, placing it outside was a deliberate "show off" by the monarch of the time), and I've looked at the pieces and they do seem to be based on it, so this change is *accurate* -- if that's all the question is you can go ahead and merge this.

But I have to wonder if most people playing the game outside of Russia will know what the heck the name is or remember it when they're using the search. Keeping it as "kremlin", while incorrect, might be more... practical for gameplay purposes? It also doesn't help that UNESCO classifies "Kremlin and Red Square" as a single entity which leads to some people thinking it *is* inside...

Gymnasiast:
I think it should be fine. For reference, I checked the Wikipedia articles of both of them, neither of which mention any potential confusion. SimCity 3000 also just refers to it as St. Basil’s cathedral. Additionally, since our search bar also looks in the object identifier, typing in "Kreml" will still show it.

We generally want to keep the descriptions accurate. We changed some descriptions of the Japanese set, for example, which described several roof and wall pieces as "Shogi", which was wrong on two counts: shogi refers to a game and it should have been "sho**j**i", and even then that only refers to the walls. So we fixed that, but you can still find the pieces by the old name:

afbeelding

Thanks for confirming this :+1:

andOlga:
Ah, well, that should be perfectly fine then if it can be searched for regardless.

PR #439, opened 2026-05-07
Drop numbers from Wacky Worlds scenery

Gymnasiast:


AaronVanGeffen (dismissed):
Hmm, I wonder to what extent this will affect identification. I see no problem dropping it from animatronics (e.g. hippo 1/2), but I wonder if it will help the situation with e.g. wall pieces. What's your take on this?

Edit: heh, didn't mean to approve this.

Gymnasiast:
Ah right, I haven’t given a lot of context here. This was the result of a conversation that Maji and I had. Only WW seems to do this (vanilla does not, neither does TT), and the numbers are often out of order to boot. It just seemed like an unnecessary complication.

ZeeMaji:
It'd also make the groups easier to maintain too, if we ever wish to change the order of items we won't have to worry about updating the names of the few items that are numerically labelled for some reason. Presently only about two dozen sets of items from solely WW use this as well so it just seems like yet another frontier fumble since they didn't even apply it to everything in the pack. (Antarctic theme has 0 numerically labelled items, South America theme only has 4.)

ZeeMaji:
From a glance at the list of diffs, to me the only one that seems like it might need a tweak is the suspension bridge start/tower pieces. I think just removing the 'side' part of the name and replacing it with 'Section' (which is already used by the cable piece from the set) would suffice.

ZeeMaji (changes requested):
The skyscraper metal set pieces from the north america theme and tudor roof extension piece from the europe theme still have numbers in their names.

ksnip_20260507-171203
ksnip_20260507-171204

ZeeMaji:
The changes in-game looked fine to me, the few objects that slipped through last time are accounted for now. Although the diffs are pretty big from the recent localisation merge.

AaronVanGeffen (approved):
Can we automatically remove the numbers for the object translations as well? Not sure if that needs to happen in the localisation repository, though.

Fine with me, either way.

PR #440, opened 2026-05-25
Scenario/park name fixes

ZeeMaji:
Fixes the capitalization of a few scenario/park names, for Six Flags Over Texas the 'Over' is capitalized in the real world park's name and was so when the scenario was made (you can check https://web.archive.org/web/20021206103010/http://www.sixflags.com/pickapark.html for proof) so this is likely an original error. The rest are just frontierisms.

Fixes the bobsled competition using bobsled in en-GB, I assume this was just an oversight since the original competitions were presumably made by hasbro of america and used en-US names as a result and this just ended up in the scenario text dumps. Now it's Bobsleigh in en-GB and Bobsled in en-US, I can move the en-US change to the localization repo if that's preferred.

I did not touch any other languages other than en-GB and the single instance of en-US for bobsled, some of them are the same as en-GB though (eg: russian) or contain the same phrase (mainly on SFOT). If changing that here is desirable then let me know. Otherwise we can let the localizers handle it in the localization repo.


Gymnasiast:
Since we’re correcting a real-life name, please also apply the SFOT correction to other languages (I’ll take care of syncing them to Localisation).

ZeeMaji:
I've went ahead and capitalized the 'Over' in all of the languages that had it in their strings, I didn't touch the rest though. (They were also all only in the BYO scenario's meta file since there were no lower case 'over's in the populated scenario's meta file.)

Gymnasiast (approved):

PR #441, opened 2026-06-06
Rename isOpaque to isTransparent

Harry-Hopkinson:
Rename isOpaque to isTransparent to reflect what the property actually stores. For https://github.com/OpenRCT2/OpenRCT2/pull/26646.


Harry-Hopkinson:
If I go through every object and flip the isOpaque value (resulting in a commit with over 200 files) it would be a huge change. I think I should change ```isOpaque``` to ```isTransparent``` which involves a lot less files.

Harry-Hopkinson:
Screenshot 2026-06-07 at 10 35 44
Screenshot 2026-06-07 at 10 36 06

Gymnasiast (approved):

PR #443, opened 2026-06-22
Add small waterfall (official.scenery_wall.wallwf16)

JanStrauss:
Adds a half height version of the waterfall wall from rct2. 

This seems to have come up a couple of times in the discord and I've recently also ran into this limitation playing around with adding rivers to the map generator.

Screenshot 2026-06-22 at 16 39 23

https://github.com/user-attachments/assets/8a5fef89-193c-43c6-9d1a-45bd87784ce4

From left to right, bottom to top:

- `rct2.scenery_wall.wallwf32`, `rct2.scenery_wall.wallwf32`
- `rct2.scenery_wall.wallwf32`, `rct2.scenery_wall.wallwf16`
- `rct2.scenery_wall.wallwf16`, `rct2.scenery_wall.wallwf16`
- `rct2.scenery_wall.wallwf32`
- `rct2.scenery_wall.wallwf16`

right side is the same mirrored.


ZeeMaji:
Any new scenery additions should be put under the OpenRCT2 official source game and folders, so the ID's namespace and the source game should be changed to "official" along with moving the object folder to `objects/official/scenery_wall`

Gymnasiast:
We discussed this and agree with adding it. We still need to test it, though.

